PR #125, opened 2014-02-06
Add Italian Language

sniperwolf:
Add Italian Language


GaryJones:
Hi @sniperwolf. Some Italian language files now exist in the [languages directory](https://github.com/Automattic/liveblog/tree/master/languages) - is there anything in those files that need improvement? (I don't think the latest `.pot` for 1.9 has been added yet)

PR #129, opened 2014-02-15
Create liveblog-it_IT.mo

misterfunk74:
.mo file for Italian localization


justnorris:
This has been open for quite a while. There is a newer version of the liveblog-it_IT.mo file in the repository. Closing.

PR #143, opened 2014-05-21
Remove font family and pixel font sizes

philipjohn:
When using this for a project today I noticed that the plugin CSS includes pixel-specific font sizes and specifies font family.

Rather, it'd be nicer if the plugin obeyed fonts and sizes from the theme and this pull request aims to achieve that.

### DIFF
--- a/css/liveblog.css
+++ b/css/liveblog.css
@@ -1,7 +1,6 @@
 /* =Container
 -------------------------------------------------------------- */
 #liveblog-container {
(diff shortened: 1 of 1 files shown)


GaryJones:
Closing this PR as it's over 10 years old and the CSS has been completely rewritten since then (now uses SCSS with the React-based frontend). The principle of using theme fonts is still valid, but would need a fresh approach with the current codebase. Thank you for the contribution\!

Issue #108, opened 2013-08-18
Deleting updated entries doesn't work


ehg:
When deleting a previously updated entry, the supposedly deleted entry will show up on a page refresh. 

To reproduce:
- Publish an entry, with text "foo"
- Edit the entry, with text "bar"
- Delete the entry
- Refresh the page

The desired outcome is that the entry does not appear, the actual outcome is that the "bar" entry is still there - and won't be deleted properly until it is deleted again.

PR #159, opened 2014-08-28
Fix: Deleting updated entries doesn't work #108

hbarroso:


rebeccahum:
This one also fixes #243

classes/class-wpcom-liveblog-entry.php
```diff
@@ -66,10 +66,13 @@ public function for_json() {
 	}
 
 	public function get_fields_for_render() {
-		$entry_id     = $this->comment->comment_ID;
-		$post_id      = $this->comment->comment_post_ID;
```
rebeccahum: I don't think `$post_id` needs to be replaced.

philipjohn:
I believe this was fixed by #309

PR #171, opened 2015-01-26
Amended CSS selector

### LINKED ISSUE #158: Wrong selector inside plupload.js
Inside plupload.js, the `#liveblog-actions` selector is used. Shouldn't the selector be `.liveblog-actions`?

Note: In liveblog-form.php the fieldset has **class** `liveblog-actions`.

henrywright:
Changed `#liveblog-actions` to `.liveblog-actions`

Fixes #158


joshbetz:
@henrywright There are two commits here. Can you fix the i158 branch to only include the CSS commit?

henrywright:
@joshbetz my apologies. I've reverted that particular commit

PR #179, opened 2015-03-12
Fixes an issue with removing replaced entries

bradaric:
The first issue was that remove_replaced_entries() in some cases removed the latest updates thus showing an older update on page refresh.

Also, get_between_timestamps() doesn't actually need to call  remove_replaced_entries() as that breaks the chain of updates for the JS code, again not showing the latest update after AJAX refresh calls.

### DIFF
--- a/classes/class-wpcom-liveblog-entry-query.php
+++ b/classes/class-wpcom-liveblog-entry-query.php
@@ -89,7 +89,7 @@ class WPCOM_Liveblog_Entry_Query {
 			}
 		}
 
-		return self::remove_replaced_entries( $entries_between );
+		return $entries_between;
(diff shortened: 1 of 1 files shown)


philipjohn:
@bradaric Hi! Did you intend to close this off, as well as #178 ? Just wanted to make sure...

bradaric:
Hi,

Yes - I realized my changes were too disruptive to the way you currently handle updates and entries as well.

My code focuses on the latest update and keeping the chain of replacements intact, while the original code writes all updates back to the initial entry and doesn't care about the replacements.

Maybe I'll fork at some other point... :|


Take care,
Alex

PR #182, opened 2015-04-22
Make the URL passed to add_query_arg safe

philipjohn:


liveblog.php
```diff
@@ -633,7 +633,7 @@ private static function add_default_plupload_settings() {
 	private static function get_entries_endpoint_url() {
 		$post_permalink = get_permalink( self::$post_id );
 		if ( false !== strpos( $post_permalink, '?p=' ) )
-			$url = add_query_arg( self::url_endpoint, '', $post_permalink ) . '='; // returns something like ?p=1&liveblog=
+			$url = add_query_arg( self::url_endpoint, '', rawurlencode( $post_permalink ) ) . '='; // returns something like ?p=1&liveblog=
```
nickdaugherty: The 3rd param to `add_query_arg()` is the url to modify (unless you pass an array as first param)...it doesn't actually need url encoded, just the individual params. In this case, it's just an empty string, so is safe as-is.

philipjohn: Damn. Then I have mis-interpreted @simonwheatley's summary of the issue.

philipjohn: I read it again and understand now

nickdaugherty: Alright cool, no worries.

PR #190, opened 2015-07-10
Extending Entries

jasonagnew:
### Overview

This pull request extends the entry system to support `#hashtags`, `/commands`, `@authors` and `:emoji:` with an autocomplete system to help speed up the process. On top of this there is also a HTML5 notification section for users. These extensions are filtered on save, for example a hashtag `#football` would be saved as `<span class="liveblog-hash term-football">football</span>` allowing easy styling. The container of the entry will also receive the same class `term-football`.

The new features are disabled by default, to enable one or more using the `wp-config.php`:

``` php
define('LIVEBLOG_FEATURES', 'commands, hashtags, emojis, authors');
```

The command system  has two inbuilt commands:

`/key`: Which defines an entry to a key event, it adds the meta key to entry `liveblog_key_entry`. In your theme you can add the shortcode `[liveblog_key_events]` where entries used with the key command will be inserted - it acts an anchor system to jump to parts of the blog.

`/alert`: Will fire off a HTML5 notification if the user has enabled them.

You can add new commands easily with a filter, the most basic command will add a class to entry, so you could do a simple  `/highlight` which would add `type-highlight` to the entry container letting you style a new background colour:

``` php
add_filter( 'liveblog_active_commands',  array( __CLASS__, 'add_highlight_command' ), 10 );


public static function add_highlight_command( $commands ) {
  $commands['highlight'] = false;
  return $commands;
}
```

A command can take both a filter called before the entry is saved, or an action that is called after it’s saved:

``` php
$commands['highlight'] = array(
  array( __CLASS__, 'filter' ),
  array( __CLASS__, 'action' ),
);
return $commands;
```

If you only need the action and not the filter you it set it to `false`:

``` php
$commands['highlight'] = array(
  false,
  array( __CLASS__, 'action' ),
);
return $commands;
```

### HTML5 Notifications
Users can subscribe to various notifications, such as author alerts, key events, as well as subscribing to as many hashtags as they'd like. This is saved using LocalStorage, so the users settings persist if they reload the page. Notifications will only appear when the window / tab  is not in focus, so the user doesn't miss any events whilst they are doing other things.

### Caveats

We have allowed commands to be sent on their own without any other text, the reasoning behind this is someone may want to add command like `/default` which uses a filter to insert a default message.

Emoji's don't use the WordPress CDN, ideally it would be better to use the standard WordPress CDN for emoji's - currently we have included: https://github.com/arvida/emoji-cheat-sheet.com

The key event system heavily uses CSS & JS instead of altering the HTML, we did this to avoid altering the entry render method. Ideally it would be good to alter the render system to pull in different templates.

### Possibilities

We were wondering if there would be any interest in moving to a Gulp based workflow for minifying, combing CSS & JS and maybe Sass, and whether using Composer to autoload/classmap files would be more helpful than `requires`.

Thanks for taking a look, love the Big Bite team.


philipjohn:
Thanks!

I've been making some inline comments with questions/feedback about specific bits of code.

There are also some general bits of feedback:

> The new features are disabled by default, to enable one or more using the wp-config.php:

This feels like an unnecessary extra step for users. Is there a reason for not including the features by default?

> The command system has two inbuilt commands:

This PR explanation is great, but it should really be plugin documentation :)

> A command can take both a filter called before the entry is saved, or an action that is called after it’s saved:

This is a nice touch, but I feel like this should be done by providing hooks like so;

```
apply_filter( "liveblog_command_{$command}_before", $arg );
do_action( "liveblog_command_{$command}_after", $arg );
```

> currently we have included: https://github.com/arvida/emoji-cheat-sheet.com

Looks like we have license incompatibility there :(

> Ideally it would be good to alter the render system to pull in different templates.

That sounds like it could be a good enhancement - want to open an issue suggesting that?

> We were wondering if there would be any interest in moving to a Gulp based workflow for minifying, combing CSS & JS and maybe Sass, and whether using Composer to autoload/classmap files would be more helpful than requires.

Okay, feel free to open an issue on those suggesting how it could be done and the benefits :)

imsimond:
I can see a few problems with the `liveblog_key_events` shortcode.

- It's currently `echo`ing its output where it should be `return`ing, as per Codex documentation. This causes real problems in any non-standard usage scenario (eg in a widget).
- I was hoping to see a few attribute options on that shortcode, perhaps allowing me to set my own title, specify a maximum number of characters to be shown per item, output full HTML or plain text...
- If we're going to keep the Key Events title, even as a default for when someone doesn't specify their own title, it needs to be wrapped in `__( 'Key Events', 'liveblog' )` markup to make it translation-ready.

I also wonder whether it should be using the full `liveblog-single-entry` template for its output. I would expect the Key Events list to be pretty concise. Just a timestamp and a single line of text, ideally trimmed to a customisable length and/or trimmed at the first line break. If you use the 'main' template, you'll probably find people having to hide most of its contents.

See for inspiration, from today's news:
- [Politico.EU - using our plugin](http://www.politico.eu/article/live-greece-votes-debt-tsipras/)
- [Guardian liveblog - ultra compact Key Events list](http://www.theguardian.com/media/live/2015/jul/16/bbc-future-set-out-in-government-green-paper-live-updates)

jasonagnew:
Hi @philipjohn 

Thanks for taking a look over it. I’ll start replying to the inline comments now.

There wasn’t any major reason for disabling the features by default, apart from avoiding the unlikely situation where a user of the plugin has made their own custom changes using `preprocess_comment`, or altering content before display — it would clash with this update. Perhaps it would make more sense if `LIVEBLOG_FEATURES` isn’t defined, then everything is enabled but it still leaves the option to easily pick and choose.

Moving to your hook idea for commands would actually simplify it a little, let me have a look into altering this.

Shame on the license issue, but it’s a good excuse to spend the time moving it over the WordPress CDN (http://s.w.org/images/core/emoji/72x72/2764.png)

Let me make a few changes to this PR to fix the above, plus adding to the plugin documentation :). And I’ll get [at]mark to make notes on we could do Gulp.

Hi @imsimond

The `echo` and not being able to set a title is a mistake on my part. I’ll update these in the PR. Regarding being able to set a different template, that was our ideal solution, but with the current render system it would be big task. We had Jon spend a day exploring this, we originally thought it would take a few hours to do, but unfortunately not. 

To work around this we used CSS to style them into list and JS to turn them into anchors. See this screenshot: http://share.agnew.co/13nOf


Cheers
Jason

classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
```diff
@@ -0,0 +1,192 @@
...
+	 * The preg replace callback for the filter.
+	 *
+	 * @param array $match
+	 * @return string
+	 */
+	public function preg_replace_callback( $match ) {
+		$term = iconv( 'UTF-8', 'ASCII//TRANSLIT', $match[2] );
+
+		if ( ! term_exists( $term, self::$taxonomy ) ) {
+			wp_insert_term( $term, self::$taxonomy );
```
imsimond: Underscores are being accepted in the creation of a hashtag but dashes aren't. There's a potential problem here, as dashes are valid characters in a term name... and I might want to pre-create a bunch of hashtags manually, ahead of my live event.

classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		return str_replace(
+			$match[1],
+			'<span class="liveblog-hash '.$this->class_prefix.$term.'">'.$term.'</span>',
```
imsimond: We seem to be using the term name for the class here, where I'd more normally be expecting the (forcible lowercase) slug.

jasonagnew:
Hi @imsimond and @philipjohn 

We’re planning to commit a few changes later today.  I’ve taken a look into having different templates for the key events shortcode and I have a solution that should be ready shortly.

Cheers
Jason

classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
ConnorVG: Hey @imsimond,

I've amended the system to allow for hyphens. The system now runs purely on slugs. For instance, if you make a new hashtag manually then you will use the slug for the hashtag. This stops the spacing issue (as hashtags can't have spaces and neither can slugs).

classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
ConnorVG: Hey @imsimond,

With my [last comment](https://github.com/Automattic/liveblog/pull/190#discussion_r34900122) — they will now be using slugs :-)

jasonagnew:
Hi @imsimond  

I’ve implemented a system to allow different templates with key events shortcode. Defining a template/type through the shortcode is not possible due to the fact the ajax calls which render new entries would never activate that shortcode - therefore I opted for 2 dropdowns in the admin panel of each post. This offers the ability to have different templates and/or formats for different liveblogs on the site.

http://share.agnew.co/15laH

Note: This is big change so it still requires some work on templates and UI. 

Cheers
Jason

jasonagnew:
Hi @imsimond 

I’ve updated the key events templates now, there is a standard list and timeline similar to the Guardian. @markgoodyear will be doing some browser testing later.

To make it easier to control I’ve added a way for developers to add their own format functions and/or templates like so:

``` php
function liveblog_add_key_event_template() {

	add_filter( 'liveblog_key_templates', 'add_template', 'liveblog-key-custom-class');
	add_filter( 'liveblog_key_formats',   'add_format' );

	function add_template( $templates ) {
		$templates['custom'] = array( '{theme}/key-events.php', 'div' );
		return $templates;
	}

	function add_format( $formats ) {
		$formats['strip-tags'] = 'new_format';
		return $formats;
	}

	function new_format( $content ) {
		$content = strip_tags( $content );
		return $content;
	}
}

add_action( 'init', 'liveblog_add_key_event_template' );
```

Cheers
Jason

css/liveblog.css
```diff
@@ -492,3 +492,170 @@ a.liveblog-form-entry::-webkit-input-placeholder {
...
+}
+
+.liveblog-key-events .liveblog-key-timeline li .link {
+	position: relative;
+	display: block;
+	padding: 0 10px 20px;
+	border-left: 1px solid #dfdfdf;
+}
+
+.liveblog-key-events .liveblog-key-timeline li .link:before {
```
imsimond: I think we need to lose the Guardian styling: too obviously ripped off.

imsimond:
Something's broken in a recent update relating to Key Events boxes in sidebar widgets. I've put a `[liveblog_key_events ... ]` shortcode in a text widget. It was updating fine with new Key Events. Now it isn't.

jasonagnew:
@imsimond for the new rendering of templates we had to alter a decent bit of the javascript so there’s a small chance your browsers cache is causing issues. Would you be able to clear your cache and test again? Should help me narrow it down a bit.

css/liveblog.css
jasonagnew: @imsimond I’ve deleted those timeline styles and pushed up an alternative style (will still require some cross browser testing).

imsimond:
I think the problem is that every shortcode on a page is being turned into `<ul id="liveblog-key-entries">` - and when a new `/key` update is sent, it's only the first one which gets updated. How likely is it that you'd have multiple 'key events' lists on the page? Admittedly, probably slim.

jasonagnew:
Ah sorry for the confusion, I’ve pushed an update to support more than one. I had been using id’s to match the main the feed, but it makes more sense using classes now to support that. Also I've updated the title setting for the shortcode so you can do `title="false"` to remove it allowing you to use widget title instead.

philipjohn:
> Perhaps it would make more sense if LIVEBLOG_FEATURES isn’t defined, then everything is enabled but it still leaves the option to easily pick and choose.

I'm in favour of just having them on, and that's it. Decisions not options ;)

README.md
```diff
@@ -40,6 +40,147 @@ If you'd like to check out the code and contribute, [join us on github](https://
...
+A command can have both a filter called before the entry is saved, or an action that is called after it’s saved:
+
+``` php
+apply_filter( "liveblog_command_{$command}_before", $arg );
+do_action( "liveblog_command_{$command}_after", $arg );
+```
+
+#### Customizing Key Events Shortcode
+
+As mentioned earlier you can add the key events section by using `[liveblog_key_events]`. If you wish to change the title from the default `Key Events` then you a title attribute.
```
philipjohn: Reading this makes me think that I need to explicitly add the shortcode before key events will work, whereas I'd expect it to work out of the box.

Also, because Liveblog takes over the content of a post, anything added into the editor of a Liveblog post is ignored - so where would the shortcode go?

README.md
```diff
@@ -40,6 +40,147 @@ If you'd like to check out the code and contribute, [join us on github](https://
...
+add_filter( 'liveblog_key_templates', 'add_template', 'liveblog-key-custom-css-class');
+
+function add_template( $templates ) {
+  $templates['custom'] = array( '{theme}/key-events.php', 'div' );
+  return $templates;
+}
+```
+There's a few things to note here:
+
+* `{theme}` is a shorthand to point to the current active theme directory, in this case we our loading template file `key-events.php`.
```
philipjohn: `{theme]` is unnecessary here, I think. We should specify that a theme should contain a `liveblog` directory into which custom templates go, and that the plugin looks there for templates *first* before looking at it's own defaults. This happens with other plugins, so should be familiar to many.

README.md
```diff
@@ -40,6 +40,147 @@ If you'd like to check out the code and contribute, [join us on github](https://
...
+* `div` is where we set the element type the wraps all entries, in the case where you wanted to built a list you would set this to `ul`.
+* `liveblog-key-custom-css-class` is a class that will added the wrapper element of the entry to help with styling, in this case that'd look like: `<div class="liveblog-key-custom-css-class">...</div>`
```
philipjohn: It feels inconsistent that the element is passed within the hooked function but the class is passed in `add_filter` - could we standardise on one of those?

README.md
```diff
@@ -40,6 +40,147 @@ If you'd like to check out the code and contribute, [join us on github](https://
...
+
+An example of a template file is:
+
+```php
+<div <?php echo $css_classes; ?> >
```
philipjohn: I'd prefer that a function (e.g. `liveblog_entry_class()`) where used here so that we can ensure the output is escaped

README.md
```diff
@@ -40,6 +40,147 @@ If you'd like to check out the code and contribute, [join us on github](https://
...
+	<a href="#liveblog-entry-<?php echo $entry_id; ?>">
```
philipjohn: Let's not have unescaped variables in examples ;)

README.md
jasonagnew: Sorry that was a mistake in the docs, its actually:

``` php
  $templates['custom'] = array( '{theme}/key-events.php', 'div', 'liveblog-key-custom-css-class' );
```
I've changed it in the docs now.

README.md
jasonagnew: I might be a little confused but `$css_classes` and `$entry_id` are both built in the render method.

https://github.com/bigbitecreative/liveblog/blob/extending-entries/classes/class-wpcom-liveblog-entry.php#L70-L93

So they should be escaped, as `$css_classes` comes from `comment_class`, and `$entry_id` isn’t a user input. I hope that makes sense?

README.md
jasonagnew: Good idea I've removed the need for `{theme}` :smile:

jasonagnew:
Cool I've removed `LIVEBLOG_FEATURES`

README.md
markgoodyear: I've updated the readme to make it less confusing to what the shortcode is for and where it might be placed in a theme.

README.md
```diff
@@ -40,6 +40,147 @@ If you'd like to check out the code and contribute, [join us on github](https://
...
+		<?php echo WPCOM_Liveblog_Entry_Key_Events::get_formatted_content( $content, $post_id ); ?>
+	</a>
+</div>
+```
+
+**Format:** Is how each entries content is filtered, there is three inbuilt formats:
```
philipjohn: I'm a little confused about formats :)

I seem to remember a screenshot showing the format being selected in the Liveblog metabox for a post. Does that mean that the same format will be applied to every entry in that Liveblog? I.e. if I choose "first sentence" then only the first sentence of every entry will show?

README.md
```diff
@@ -40,6 +40,147 @@ If you'd like to check out the code and contribute, [join us on github](https://
...
+
+Selecting which template or format to use for liveblog happens in the admin panel on the edit of page of the post:
+
+![Key Events Admin Options](http://share.agnew.co/Gyai+)
+
+#### Managing Hashtags
+Hashtags are manageable in the admin area under Posts you will see menu Hashtags. Please note that the slug is used not the name.
+
+#### HTML5 Notifications
+The notification checkbox for users will only appear if there browser supports Notification API, the checkbox looks like so:
```
philipjohn: Typo on "there"?

README.md
philipjohn: Basically, `$entry_id` is a variable, and so it's value is errm variable, and therefore can't be trusted, so should be escaped. Anything where we aren't certain what will be echoed (hard-coded strings/integers) should be escaped as late as possible to avoid anything (including hooks) changing those variables to something undesirable.

https://vip.wordpress.com/documentation/best-practices/security/validating-sanitizing-escaping/#always-escape-late

js/liveblog-publisher.js
```diff
@@ -119,15 +113,13 @@
 		entry_inputhandler_textarea: function () {
 			var html = this.$textarea.val();
 			if ( ! html ) {
-				html = '<p><br></p>'; // the BR is needed to make sure browsers will land inside the <p>
+				// html = '<p><br></p>'; // the BR is needed to make sure browsers will land inside the <p>
```
philipjohn: If this is no longer needed, can we strip it out?

js/liveblog.js
```diff
@@ -382,6 +398,147 @@ window.liveblog = window.liveblog || {};
...
+	};
+
+	liveblog.get_notification_entry_text = function( $entry ) {
+		var original_content, $original_content, entry_text;
+
+		// Grab original content from data-attr
+		original_content = $entry.find('.liveblog-entry-text').data('original-content');
+
+		// Convert to jQuery object and wrap it so we can manipulate it
+		$original_content = $('<div/>').html(original_content);
```
philipjohn: This is kinda unsafe, because we can't guarantee `original_content` isn't nasty. See https://vip.wordpress.com/2015/03/25/preventing-xss-in-javascript/

philipjohn:
I've done a full once-over of the code now, and wanted to share my general findings in addition to specific feedback;

* the structure is really nice, especially the extensive use of hooks
* documentation is slim, I feel, and needs beefing up a lot
* inline documentation is often very vague and it's really hard to understand the code
* there's a fair bit of unescaped output, and I'd prefer that when HTML is being built any variables are escaped inline too
* looks like there's some inconsistency in following the WordPress coding standards, but nothing major
* thank you for being so responsive to and collaborative on our piecemeal feedback

README.md
jasonagnew: Okay makes sense. Thanks for explaining. I’ve updated those to be escaped and altered the css from:

``` php
<div <?php echo $css_classes; ?> >
```
To:
``` php
<div class="<?php echo esc_attr( $css_classes ); ?>" >
```

I also updated some of the original templates to include escaping.

js/liveblog.js
markgoodyear: I've tested performing XSS attacks, and everything entered into the text box gets filtered before saving, so anything nasty gets stripped from the `data-original-content` attribute, as well as the DOM. We can change it to grab the DOM directly so we don't need to go through the `.html()` method, something like this:

```
// Grab the liveblog entry text, clone to variable
$original_content = $entry.find('.liveblog-entry-text').clone();

// Remove command spans
$original_content.find('.liveblog-command').remove();

// Convert to text string
entry_text = $original_content.text();
```

jasonagnew:
Thanks @philipjohn for taking the time to review our code, we really appreciate it  - we’ve replied to some of the comments and we’ll address your concerns on documentation and inline comments before 6pm today :smile:

README.md
jasonagnew: Yip that's fixed now.

js/liveblog-publisher.js
jasonagnew: Yip that's removed now.

README.md
jasonagnew: I've added a little more info to the docs on formats so it makes a bit more sense now:

Formats add an extra level of control to how the key events section looks. If using the `timeline` template with `first sentence` and the following is entered into the new entry box:
```
New iPad announced. With plan to ship next month, pre-orders starting 23rd September retailing at $499 for 16gb $599 32gb. /key
```

The main feed and notification would show the full text, and the key events section would only show:
```
New iPad announced
```

jasonagnew:
Hi @philipjohn we’ve beefed up the documentation and included inline comments as promised.

js/liveblog.js
philipjohn: That looks like a good solution, yeah!

markgoodyear: @philipjohn I've pushed an update for this :)

PR #195, opened 2015-08-12
Add additional escaping to single entry template

elliott-stocks:
There's a few places with doesn't have any escaping for front-end output. This fixes that!


philipjohn:
Thanks @elliott-stocks - could I ask you to do a new PR for merging into develop please, and I'll get that merged shortly?

PR #198, opened 2015-08-21
Fixes multiple hashtag terms being added on frontend

pkevan:
Check for wp_error when inserting term


philipjohn:
Removed completely in #200

PR #201, opened 2015-08-21
Changing preg to not break on mulitple words separated by hypens

pkevan:
catches hypens now, rather than just first word


philipjohn:
Thanks @pkevan that's tested and on wpcom

PR #209, opened 2015-11-10
Key Events compatible lazyloading

tfrommen:
Hi there.

This pull request realizes user-driven lazyloading of Liveblog entries:
* displays a (filterable) number of entries on page load;
* fetch other entries via AJAX;
* on button click, display more entries;
* fully capable of working with (not yet displayed) Key Events;
* filters:
  * `liveblog_enable_lazyloader`, enable/disable lazyloading;
  * `liveblog_number_of_entries`, number of entries (for initial display, and lazyloading);
* disables _Lazyload Liveblog Entries_ plugin as it would collide with the built-in lazyloading.

### DIFF
--- a/classes/class-wpcom-liveblog-entry-query.php
+++ b/classes/class-wpcom-liveblog-entry-query.php
@@ -139,4 +139,35 @@ class WPCOM_Liveblog_Entry_Query {
 
 		return $result;
 	}
+
+	/**
+	 * Returns the Liveblog entries between the two given (optional) timestamps.
+	 *
+	 * @param int $max_timestamp Maximum timestamp for the Liveblog entries.
+	 * @param int $min_timestamp Minimum timestamp for the Liveblog entries.
+	 *
+	 * @return WPCOM_Liveblog_Entry[]
+	 */
+	public function get_for_lazyloading( $max_timestamp, $min_timestamp ) {
+
+		$entries = $this->get_all();
+		if ( ! $entries ) {
+			return array();
+		}
+
+		if ( $max_timestamp ) {
+			foreach ( $entries as $key => $entry ) {
+				$timestamp = $entry->get_timestamp();
+
+				if (
+					( $max_timestamp && $timestamp >= $max_timestamp )
+					|| ( $min_timestamp && $timestamp <= $min_timestamp )
+				) {
+					unset( $entries[ $key ] );
+				}
+			}
+		}
+
+		return $entries;
+	}
 }
--- a/classes/class-wpcom-liveblog-lazyloader.php
+++ b/classes/class-wpcom-liveblog-lazyloader.php
@@ -57,7 +57,7 @@ class WPCOM_Liveblog_Lazyloader {
 	public static function late_load() {
 
 		if ( has_action( 'init', 'Lazyload_Liveblog_Entries' ) ) {
-			if ( is_admin() ) {
+			if ( is_admin() && current_user_can( 'activate_plugins' ) ) {
 				add_action( 'admin_notices', array( __CLASS__, 'admin_notices' ) );
 			}
 
--- a/classes/class-wpcom-liveblog-lazyloader.php
+++ b/classes/class-wpcom-liveblog-lazyloader.php
@@ -138,7 +138,7 @@ class WPCOM_Liveblog_Lazyloader {
 			$number = $number_of_default_entries;
 		}
 
-		$args['number'] = $number;
+		$args['number'] = (int) $number;
 
 		return $args;
 	}
@@ -163,7 +163,7 @@ class WPCOM_Liveblog_Lazyloader {
 		wp_enqueue_script( $handle, plugins_url( $path, $plugin_path ), array( 'liveblog' ), filemtime( $temp ), true );
 		wp_localize_script( $handle, 'liveblogLazyloaderSettings', array(
 			'loadMoreButtonText' => esc_html__( 'Load more entries&hellip;', 'liveblog' ),
-			'numberOfEntries'    => self::$number_of_entries,
+			'numberOfEntries'    => (int) self::$number_of_entries,
 		) );
 	}
 }
--- a/classes/class-wpcom-liveblog-lazyloader.php
+++ b/classes/class-wpcom-liveblog-lazyloader.php
@@ -84,7 +84,11 @@ class WPCOM_Liveblog_Lazyloader {
 		 *
 		 * @param int $number_of_entries Number of Liveblog entries.
 		 */
-		self::$number_of_entries = (int) apply_filters( 'liveblog_number_of_entries', self::$number_of_entries );
+		$number_of_entries = (int) apply_filters( 'liveblog_number_of_entries', self::$number_of_entries );
+		// TODO: Limit number of entries to some not yet defined maximum value.
+		if ( $number_of_entries > 0 ) {
+			self::$number_of_entries = $number_of_entries;
+		}
 
 		add_action( 'wp_enqueue_scripts', array( __CLASS__, 'enqueue_script' ) );
 
--- a/classes/class-wpcom-liveblog-lazyloader.php
+++ b/classes/class-wpcom-liveblog-lazyloader.php
@@ -137,7 +137,7 @@ class WPCOM_Liveblog_Lazyloader {
 		 *
 		 * @param int $number_of_default_entries Number of initially displayed Liveblog entries.
 		 */
-		$number = (int) apply_filters( 'number_of_default_entries', 10 );
+		$number = (int) apply_filters( 'number_of_default_entries', $number_of_default_entries );
 		if ( $number < 0 ) {
 			$number = $number_of_default_entries;
 		}
--- a/classes/class-wpcom-liveblog-lazyloader.php
+++ b/classes/class-wpcom-liveblog-lazyloader.php
@@ -80,14 +80,13 @@ class WPCOM_Liveblog_Lazyloader {
 		}
 
 		/**
-		 * Filters the number of Liveblog entries used both for initial display, and lazyloading.
+		 * Filters the number of Liveblog entries used for lazyloading.
 		 *
 		 * @param int $number_of_entries Number of Liveblog entries.
 		 */
-		$number_of_entries = (int) apply_filters( 'liveblog_number_of_entries', self::$number_of_entries );
-		// TODO: Limit number of entries to some not yet defined maximum value.
+		$number_of_entries = min( apply_filters( 'liveblog_number_of_entries', self::$number_of_entries ), 100 );
 		if ( $number_of_entries > 0 ) {
-			self::$number_of_entries = $number_of_entries;
+			self::$number_of_entries = (int) $number_of_entries;
 		}
 
 		add_action( 'wp_enqueue_scripts', array( __CLASS__, 'enqueue_script' ) );
@@ -137,7 +136,7 @@ class WPCOM_Liveblog_Lazyloader {
 		 *
 		 * @param int $number_of_default_entries Number of initially displayed Liveblog entries.
 		 */
-		$number = (int) apply_filters( 'number_of_default_entries', $number_of_default_entries );
+		$number = apply_filters( 'number_of_default_entries', $number_of_default_entries );
 		if ( $number < 0 ) {
 			$number = $number_of_default_entries;
 		}
--- a/classes/class-wpcom-liveblog-lazyloader.php
+++ b/classes/class-wpcom-liveblog-lazyloader.php
@@ -166,8 +166,8 @@ class WPCOM_Liveblog_Lazyloader {
 		$temp = plugin_dir_path( $plugin_path ) . $path;
 		wp_enqueue_script( $handle, plugins_url( $path, $plugin_path ), array( 'liveblog' ), filemtime( $temp ), true );
 		wp_localize_script( $handle, 'liveblogLazyloaderSettings', array(
-			'loadMoreButtonText' => esc_html__( 'Load more entries&hellip;', 'liveblog' ),
-			'numberOfEntries'    => (int) self::$number_of_entries,
+			'loadMoreText'    => __( 'Load more entries&hellip;', 'liveblog' ),
+			'numberOfEntries' => (int) self::$number_of_entries,
 		) );
 	}
 }
--- a/classes/class-wpcom-liveblog-lazyloader.php
+++ b/classes/class-wpcom-liveblog-lazyloader.php
@@ -10,29 +10,100 @@ class WPCOM_Liveblog_Lazyloader {
 	/**
 	 * @var bool
 	 */
-	private static $enabled = true;
+	private static $enabled;
 
 	/**
 	 * @var int
 	 */
-	protected static $number_of_entries = 5;
+	private static $number_of_default_entries;
 
 	/**
-	 * Checks if the lazyload feature is enabled.
+	 * @var int
+	 */
+	private static $number_of_entries;
+
+	/**
+	 * Checks if lazyloading is enabled.
 	 *
 	 * @return bool
 	 */
 	public static function is_enabled() {
 
+		if ( ! isset( self::$enabled ) ) {
+			/**
+			 * Enables/Disables lazyloading for Liveblog entries.
+			 *
+			 * @param bool $enabled Enable lazyloading for Liveblog entries?
+			 */
+			self::$enabled = (bool) apply_filters( 'liveblog_enable_lazyloader', true );
+			if ( self::$enabled && self::is_robot() ) {
+				// No lazyloading for robots.
+				self::$enabled = false;
+			}
+		}
+
 		return self::$enabled;
 	}
 
 	/**
-	 * Returns the number of Liveblog entries used both for initial display, and lazyloading.
+	 * Checks if the current user is a robot.
+	 *
+	 * @return bool
+	 */
+	private static function is_robot() {
+
+		if ( ! isset( $_SERVER['HTTP_USER_AGENT'] ) ) {
+			return false;
+		}
+
+		return (bool) preg_match( '/bot|crawl|slurp|spider/i', $_SERVER['HTTP_USER_AGENT'] );
+	}
+
+	/**
+	 * Returns the number of initially displayed Liveblog entries.
+	 *
+	 * @return int
+	 */
+	private static function get_number_of_default_entries() {
+
+		if ( ! isset( self::$number_of_default_entries ) ) {
+			self::$number_of_default_entries = 5;
+
+			/**
+			 * Filters the number of initially displayed Liveblog entries.
+			 *
+			 * @param int $number_of_default_entries Number of initially displayed Liveblog entries.
+			 */
+			$number = (int) apply_filters( 'liveblog_number_of_default_entries', self::$number_of_default_entries );
+			if ( $number >= 0 ) {
+				self::$number_of_default_entries = $number;
+			}
+		}
+
+		return self::$number_of_default_entries;
+	}
+
+	/**
+	 * Returns the number of Liveblog entries used for lazyloading.
 	 *
 	 * @return int
 	 */
 	public static function get_number_of_entries() {
+
+		if ( ! isset( self::$number_of_entries ) ) {
+			self::$number_of_entries = 5;
+
+			/**
+			 * Filters the number of Liveblog entries used for lazyloading.
+			 *
+			 * @param int $number_of_entries Number of Liveblog entries.
+			 */
+			$number = (int) apply_filters( 'liveblog_number_of_entries', self::$number_of_entries );
+			if ( $number > 0 ) {
+				// Limit the number of Liveblog entries used for lazyloading to 100.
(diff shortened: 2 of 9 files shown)


joshbetz:
Note: We need to ensure all entries are still output to crawlers like googlebot so that liveblogs are still indexed correctly.

philipjohn:
Thanks folks! Some feedback for ya, and a possible bug....

/classes/class-wpcom-liveblog-lazyloader.php:83 - we should probably have a hard limit on this to prevent silliness
/classes/class-wpcom-liveblog-lazyloader.php:111 - it looks like by default the initial Liveblog will only show 5 entries because of this code, which seems too few. Maybe we need a separate filter which falls back to that value?
/classes/class-wpcom-liveblog-lazyloader.php:136 - should be escaped to ensure it's an int
/classes/class-wpcom-liveblog-lazyloader.php:137 - should be escaped to make sure it's a URL
/js/liveblog-lazyloader.js:26 - does this just remove the button on failure? What else happens?
/js/liveblog-lazyloader.js:80 - unescaped output (using `.html()`)

Found a bug: "Load more" button remains when jumping to a key event

button-remains

Steps to reproduce:

* Have a liveblog with lots of entries, including a key event early on
* Load the liveblog in a browser (there should be a key event beyond the first five shown)
* Click on a key event that is currently hidden
* Liveblog will load more events, including the key one
* The button remains in the middle, as in the screenshot above

I realise this could be deliberate, because there are entries in-between, but it looks janky. Maybe the button text needs to be different and/or better styling?

It'd be good to get some thoughts on Josh's comment too - I haven't tested that yet.

philipjohn:
One more, actually:

/liveblog.php:482 - It looks like here the plugin loads all of the liveblog entries everytime the Ajax method is called, even if all we need is the next five. Am I reading that right?

tfrommen:
Hi there.

Here's some feedback on your feedback.

---

>[/classes/class-wpcom-liveblog-lazyloader.php:83](https://github.com/inpsyde/liveblog/blob/7478e0417a6d675323aae29ffcab1713a10c9546/classes/class-wpcom-liveblog-lazyloader.php#L83) - we should probably have a hard limit on this to prevent silliness

In case you are talking about making sure to at least have a positive number, yes, that makes sense. We can add this. Or did you mean something else?

---

>[/classes/class-wpcom-liveblog-lazyloader.php:111](https://github.com/inpsyde/liveblog/blob/7478e0417a6d675323aae29ffcab1713a10c9546/classes/class-wpcom-liveblog-lazyloader.php#L111) - it looks like by default the initial Liveblog will only show 5 entries because of this code, which seems too few. Maybe we need a separate filter which falls back to that value?

Yes, both the number of initally displayed entries and the number of lazyloaded (or better: _lazydisplayed_) entries is set by this value. We can, of course, split this into two individually, filterable values, if you wish.

---

>[/classes/class-wpcom-liveblog-lazyloader.php:136](https://github.com/inpsyde/liveblog/blob/7478e0417a6d675323aae29ffcab1713a10c9546/classes/class-wpcom-liveblog-lazyloader.php#L136) - should be escaped to ensure it's an int

Do we really need to cast (again)? The property is protected, and there is no public setter. It is set in the `late_load` method, where it is filtered and casted.

---

>[/classes/class-wpcom-liveblog-lazyloader.php:137](https://github.com/inpsyde/liveblog/blob/7478e0417a6d675323aae29ffcab1713a10c9546/classes/class-wpcom-liveblog-lazyloader.php#L137) - should be escaped to make sure it's a URL

How about removing this, and use the value already defined in the globally accessible `liveblog_settings.endpoint_url` property? This seems to be the right way of doing this.

---

>[/js/liveblog-lazyloader.js:26](https://github.com/inpsyde/liveblog/blob/7478e0417a6d675323aae29ffcab1713a10c9546/js/liveblog-lazyloader.js#L26) - does this just remove the button on failure? What else happens?

Well, this means there are no more entries to be lazyloaded&mdash;so no button.

---

>[/js/liveblog-lazyloader.js:80](https://github.com/inpsyde/liveblog/blob/7478e0417a6d675323aae29ffcab1713a10c9546/js/liveblog-lazyloader.js#L80) - unescaped output (using `.html()`)

The `settings.loadMoreButtonText` property holds this, which was passed from PHP to JS:
```php
esc_html__( 'Load more entries&hellip;', 'liveblog' )
```

Is there still anything to be escaped?

---

>[/liveblog.php:482](https://github.com/inpsyde/liveblog/blob/7478e0417a6d675323aae29ffcab1713a10c9546/liveblogphp.js#L482) - It looks like here the plugin loads all of the liveblog entries everytime the Ajax method is called, even if all we need is the next five. Am I reading that right?

Yes, and no. Every lazyloading AJAX request loads all entries, that's right. There is, however, only a single AJAX request (triggered in the `initialize` JS function).

We decided to **not** send several AJAX requests, wait for response, and then display it. Instead, we send a single request for all entries (minus the ones that are already displayed), and then batch-by-batch display them, just like the user wants.

The reason for this is that we don't want to present the user links to not yet displayed Key Event entries, and then have her wait for them to display&mdash;we discussed this with @imsimond a while ago.

---

>Found a bug: "Load more" button remains when jumping to a key event [...] I realise this could be deliberate, because there are entries in-between, but it looks janky. Maybe the button text needs to be different and/or better styling?

As you already pointed out, this is no bug, it's _by design_.

Suppose you have a Liveblog with let's say 200 entries. And many of these are content-heavy (embeds, large images, other rich content). But only some of the very first entries are Key Events.

When clicking a link to one of these Key events, I guess you wouldn't really want to load everything _above_, including all the rich content&mdash;and consequently make no real use of lazyloading/lazydisplaying at all. That's why there are only up to _n_ entries displayed when you click a Key Event link, starting with the Key entry itself. And if there are _missing_ entries before the new ones another button is added.

There is no individual styling for now, yes, so it's currently up to the theme how the load-more buttons are being displayed.

---

>It'd be good to get some thoughts on Josh's comment too - I haven't tested that yet.

Sorry, I understood this more as a reminder for you guys.

One would simply turn off lazyloading at all for crawlers, right?

I'm thinking of a simple check like this:
```php
if (
	isset( $_SERVER['HTTP_USER_AGENT'] )
	&& preg_match( '/bot|crawl|slurp|spider/i', $_SERVER['HTTP_USER_AGENT'] )
) {
	// No lazyloading for you.
}
```

Are you okay with this?

---

We look forward to your response.

Thanks.

imsimond:
I don't like the use of a `<button>` where we have a break in the timeline. The 'convention' (?) for such things is something like:

ce5j2lmweaa-xtp

as seen in many Twitter apps, when you return after an extended period away. So I think this needs to be a clickable link, which can be styled as a full-width `display:block`, allowing a designer to add 'jagged edge' styling above and below.

tfrommen:
@imsimond don't you really want to *use* a `<button />`, or don't you like that the button is displayed inline?

In fact, we shouldn't use an `a` here, because clicking the element neither brings the user to some external site, nor does it reference some on-page element&mdash;it triggers an action.

My suggestion is to define some appropriate styling for the `<button />` instead, including `display: block`.

philipjohn:
> In case you are talking about making sure to at least have a positive number, yes, that makes sense. We can add this. Or did you mean something else?

We certainly want to avoid non-positive values here, yes, but this is really about what kind of load we're generating - both at the server and client side. For example, while `-1` would obviously be bad, so might `1000`, `500` or even `100`. We need to protect sites from having queries that are greedier than we absolutely need, really.

> We can, of course, split this into two individually, filterable values, if you wish.

I think that'd be a good improvement - I can imagine folks asking for this so let's head that off now.

> Do we really need to cast (again)?

Short answer: yes. Long answer: This comes down to the principle of late escaping (https://vip.wordpress.com/documentation/best-practices/security/validating-sanitizing-escaping/#always-escape-late). While the output is escaped, it's not escaped at the latest possible point. That means that a future code change could introduce Bad Things(TM) in a way that isn't immediately noticeable from reviewing a diff of changes. Late escaping *helps* to avoid this.

> How about removing this, and use the value already defined in the globally accessible liveblog_settings.endpoint_url property? This seems to be the right way of doing this.

Great suggestion!

> Is there still anything to be escaped?

See above point re late escaping.

> Every lazyloading AJAX request loads all entries, that's right.

Okay, I feel like we should only be pulling the entries we *need* to show. There is a method there already based on timestamps that we can use. I understand the plugin already loads all entries in one go, but that's still sub-optimal. We'd probably want a higher limit that 5, too, so that fewer AJAX requests are made. We can also cache the queries results to keep the DB load low, too.

> As you already pointed out, this is no bug, it's by design.

Okay, I can see that makes sense. I worry that out of the box it's not a great UX, and that we shouldn't rely that much on the theme to make it pretty. What more can we do with this? I'm not so bothered about the underlying HTML as much as the experience, here.

imsimond:
About the use of `<button>`: that's a good response. And you're right, of course: it's possible to achieve the same result using `<button>` with the same CSS. Should we have extra classes on the various buttons, to indicate where on the page they are appearing... and thus to help designers style them differently? I might want a conventional button at the bottom, but a 'jagged edge' in the middle of a stream.

dnaber-de:
Hi there, I'm David and I reviewed Thorsten's code before we started the pull request. So I'm familiar with the task. (Thorsten is already on his way to WordCamp Berlin this weekend.)

I would like to summarize all the issues we got in this thread so we have a nice overview.

### 1. Visibility for search engines
(By @joshbetz)
Thorsten suggested a solution:
> One would simply turn off lazyloading at all for crawlers, right?
>
> I'm thinking of a simple check like this:
> ```
if ( isset( $_SERVER['HTTP_USER_AGENT'] )
&& preg_match( '/bot|crawl|slurp|spider/i', $_SERVER['HTTP_USER_AGENT'] )
) {
// No lazyloading for you.
}
```
> Are you okay with this?

### 2. Filterable configurations

State of the discussion:

We split the number of _initially displayed_ entries and the number of entries _displayed on demand_ into two separate, filterable values:

 * `liveblog_number_of_entries`
 * `liveblog_number_of_default_entries`

philipjohn wanted us to sanitize the value. We sure can do this but keep in mind that we're talking about a filterable value not a user input. Do you really want to patronize developers here? In my opinion, if a developer wants to set a value to 1000, he might have a good reason for this.

### 3. Late escaping
We're taking already care of this.

### 4. Concept of how many entries are loaded and when
State of the discussion:

Our current solution is by design for some reasons, Thorsten pointed out earlier. Just to sum up: the idea is that payload is less expensive than HTTP-Request. Sure, this approach _has_ its limits.

So either we go the way, philipjohn suggested and load only the amount of entries we actual display.

Or we combine both ways by fetching, let's say, 20 entries per request but displaying only five. The benefit would be a better UX (delay on only every fourth lazy load), less HTTP requests (means also, less cache entries gets invalidated when one more liveblog entry is added). Both values would sure be filterable. 

However, in both cases we have to implement a solution to search for key-events which is no problem but more complex than the current solution.

## 5. Default styling of the button
We keep the button markup but we provide a default styling in order to get a better UX.

I think these should be all the points we're discussing right now.

philipjohn:
Thanks for the summary, @dnaber-de that's useful

1. Sorry, yes, that solution looks good - we should definitely include that in the plugin

2. Filterable configurations

> Do you really want to patronize developers here?

This is about making sure that sites running Liveblog remain as secure and performant. It's easy for us as developers to write code we think will be okay, only for it to produce unexpected issues and/or bugs. Ensuring we're escaping or limiting values in this way helps to protect sites from those incidents that have the potential take large sites down.

> In my opinion, if a developer wants to set a value to 1000, he might have a good reason for this.

That just isn't going to fly on high-traffic sites I'm afraid, where we need to be extra conscious about what loads we're generating. This is especially true for Liveblog which is used for high-traffic events.

3. Late escaping

Just to clarify - do you have changes incoming for these?

4. Concept of how many entries are loaded and when

This one is a bit subjective. Liveblog already loads all entries by default (which isn't great, imo) so having the lazyloading do the same *should* be fine. I'm inclined to believe that we shouldn't just do that though. Your concept of loading everything so that the UI makes a huge amount of sense, and I appreciate that approach. Your suggestion of 20 entries per request is a good one, that seems like a happy medium.

5. Default styling of the button

This one is fairly minor in the grand scheme of things, but I think if we can just have the "button" match the width of the entries, with a bit more padding top and bottom to separate it out and distinguish it that'd be good for 'out of the box'.

tfrommen:
Hello.

I just pushed a number of commits doing the following:

* use URL defined in `liveblog_settings` object instead of setting it up again;
* show the admin notice about deactivating _Lazyloading Liveblog Entries_ only for users who can actually deactivate plugins;
* turn off lazyloading for robots so that the Liveblog will be correctly indexed;
* use a separate (filterable) value for the number initially displayed Liveblog entries (independent of the number that is to be lazyloaded);
* perform late escaping;
* check return values of the number filters for positive values;
* add basic styling for the load more buttons;
* escape translated content for Javascript by using `.text()`.

What I need now is a specific value for your desired [maximum number of entries to be lazyloaded](https://github.com/inpsyde/liveblog/blob/69c95e5c2cfaf7ba6ed62949e24a9ad00ea6d501/classes/class-wpcom-liveblog-lazyloader.php#L88). You already mentioned 1000, 500, and 100&mdash;so which one is it?

---

Then, there is still the large issue concerning the actual lazyloading.

We will write up a short yet detailed summary and comment here again.

tfrommen:
# Our Reasons for the Current Lazyloading Implementation

Instead of firing a (small) AJAX request every time the user triggers lazyloading, we prefer to perform a single (large) AJAX request on page load.
Here are the reasons for our decision.

Currently, the websites that use Liveblog, in general, don't use any lazyloading. This means that a user who visits a Liveblog for the first time loads the following:

* website markup;
* images;
* script files (CSS and JS);
* possible embeds;
* possible other external resources (e.g., fonts);
* and the Liveblog entries.

What we do now is reduce the Liveblog entries to a minimum (by default: 10), and then load the rest of them in a separate AJAX request. Without the website's markup, images, script files etc. Only the HTML for the Liveblog entries. Plus some other information (ID, type) not worth mentioning.
Unless defined otherwise, AJAX requests do not have any timeout. Unless defined otherwise, GET requests are handled asynchronously, without affecting what's going on in the browser.
This means, that iff there is a website that will run into performance issues due to the fact that we load all entries in one go, that same website **cannot** function right now, no matter what. It doesn't get loaded completely now, and is displayed in some broken way. With lazyloading (and a performance issue), the website will be displayed just fine&mdash;only the Liveblog entries won't be displayed.

To be honest, however, we really can't think of any situation where you could run into such performance issues.

We do load all entries (for a given post ID) in one go, yes. We don't pose any conditions on the SQL query, though. We fetch every single Liveblog entry for a given ID, sorted by comment date (in the correct order, not like you do in the `between_timestamps` function), remove the first _n_ entries, then return it. One SQL query. No need to ask the database anything else.

The whole logic after having fetched the Livelog entries happens on the client side. A fast O(n) search in a (two-dimensional) array. This could even be improved by using the ID of the first Liveblog entry in each "entry set" as its according set ID (currently we are using an incremented index). There is no need for any other AJAX/HTTP request, and thus no waiting time (other than the almost neglibile time for finding the Liveblog entries that are to be displayed after clicking a load-more button).

In general, using one (or a few) requests with larger payload performs by far higher than using several smaller requests. Redundant headers, and either by far more time needed, or (when requesting in parallel) it is necessary to re-construct the individual responses to a single collection.

This also means that there is only a single cache entry, and thus only a single **invalid** cache entry&mdash;and not several invalid cache entries for every new request.

Of course, there is no limit for the content of a Liveblog entry. In real life, these entries wouldn't be too large, right?
Let's assume a text consisting of about 250 words. This is between 1.5 and 3 kB.
Now suppose there is an event with a lot of entries to be published. How long would the duration of such an event be? How frequent would new entries be posted? Even if this would be a one-day event, and even if there would be a new Liveblog entry every minute, this would be a total of about 1500 entries. With the above average size, this would be something between 2.5 and 3 MB. Even this shouldn't be a problem, right? And these numbers are quite high and unrealistic. Or do you know of Liveblog use cases of a totally different dimension?

---

Another important thing is that in our call with @imsimond, one requirement had been to serve a user who clicks on a link to a Key Event entry the according _result_ as fast as possible. In case the according entry is not yet displayed on the website, this takes time. Quite some time.

No matter if we are requesting exactly the same number of entries that we also want to display, or if we request more entries, and then display just a fraction of them&mdash;there are several (possible) problems with either approach. Who says the user wants the next _n_ entries after having clicked on a Key Event link? Maybe the user is interested in the Key Events only. Which means that we have to send multiple requests, and then "throw away" parts of their response. Also, multiple Key Events with more ordinary entries in between than the number of entries requested lead to even more problems.

----

So, apart from the fact that we already have an implementation that works, and apart from the fact that we sent you a proposal including a description of what we were planning to implement (which you read and accepted)&mdash;any changes to what we already have takes additional time and makes things more complicated than necessary, and (in our opinion) does not improve the overall performance.

If you are really not convinced yet, then we can do it any other way, no problem at all. Please let us know the exact procedure how you like us to do it. Maybe we can also discuss it on the phone.

Thanks a lot for your attention.

---

Please also don't forget giving us a specific value for your desired [maximum number of entries to be lazyloaded](https://github.com/inpsyde/liveblog/blob/69c95e5c2cfaf7ba6ed62949e24a9ad00ea6d501/classes/class-wpcom-liveblog-lazyloader.php#L88). Thanks.

imsimond:
Thanks Thorsten. It's a tricky balancing act.

Fundamentally you are right, of course. If we download the full set of updates in a single large payload, we can deliver the fastest possible response. Downloading the full payload will be fine for most users. But the payload may be too large for some users, for reasons beyond our control. The user may be on a mobile device with limited memory; or the page may already be making significant demands on the browser.

To mitigate against that, we will need some degree of incremental loading. But @philipjohn has some ideas for a compromise solution..?

philipjohn:
Your reasoning for the approach is sound, and makes a lot of sense. We're essentially discussing a trade-off here between client-side and server-side performance, we just need to figure out the right balance.

So... one thing we could consider to still get the benefit of your approach is to load more than requested. So, our `liveblog_number_of_entries` governs how many entries will be *shown* when clicking "load more" but we could make the query behind that always return 20/50/100 entries and store them locally. E.g. if our query limit were 100 and  `liveblog_number_of_entries` were set to 10, it would take 10 clicks but only one AJAX request to load 100 entries. We get the benefit of fast rendering on the front end but protect ourselves against the potential for performance-impacting requests on the backend.

tfrommen:
As specified by @imsimond, the maximum number of Liveblog entries that can be lazyloaded/-displayed in one go is now set to 100.

I understand the high-level description of what you would like to have is loading a number of entries and then display only a fraction of them. Having some entries in reserve, however, only is of advantage if the user actually wants to display _n_ entries that all are in our reserve. If not, we have to fire a new AJAX request and handle the response. And the Key Events make all that very complicated and complex.

What if we have entries in reserve, but not the ones the user wants? What do we do with the entries in reserve once we have the response of the new AJAX request (for the entries the user is interested in, plus several entries after that)? If we were to _throw away_ the old entries, we wasted time. If we were to keep them, we somehow had to attach information to them. This either leads to a more complex data structure, or a more complex (i.e., lower performing) search for entries, or both.

So, a possible implementation could look as follows:

* on page load, fetch the next _n_ entries, where _n_ could be related to `liveblog_number_of_entries`, or defined independently (of course it has to be at least equal to this, if not greater);
* as long as there are `liveblog_number_of_entries` entries left (in the according JavaScript array), display them;
* as soon as there are less than `liveblog_number_of_entries` entries left, fetch the next _n_ entries;
* when the user clicks on a Key Event entry:
  * if the according element is already present, nothing to do here (user jumps to the element);
  * if the element is not yet displayed, search in the JavaScript array:
    * if the according entry is present **and** if there are at least `liveblog_number_of_entries` - 1 number of entries after it, display the entries;
	* if the according entry is present **but** if there are not enough entries after it, send an AJAX request first, append the response, then display the entries.

If you are okay with this, I'm happy to proceed.

Thanks in advance for your attention and feedback.

philipjohn:
That sounds great @tfrommen ! :)

tfrommen:
Hello,

we're done with the refactoring of the lazyloading implementation, as you can see in this pull request.

Due to some (potential) problems with Key Event entries at various places in a _gap_, we chose a different implementation than stated in my last comment. We don't fetch more entries than we display. Instead we pre-fetch the next _n_ entries directly after having displayed some entries. Thus, the user always renders entries that have already been loaded into JavaScript. Only klicking on a Key Event link might lead to a short waiting time - in case the desired entry has not yet been fetched, that is.

So, all in all, this should be pretty much exactly what we're aiming for with our compromise solution, right?

philipjohn:
Thanks @tfrommen - I've been through those changes and we just have one possible issue with the use of `HTTP_USER_AGENT` and I think we're done!

tfrommen:
Hi,

I just pushed a newer version of the `is_robot` method that takes care of possible caching plugins. I therefore had to refactor the code a little bit.

philipjohn:
Great! Drum roll please...

tfrommen:
Drum Roll

PR #221, opened 2015-12-10
Return false if there is no post global

### LINKED ISSUE #220: Notice: Trying to get property of non-object in /srv/www/master/html/wp-content/plugins/liveblog/liveblog.php on line 381
The above notice was being thrown for me when accessing the Plugins menu in Network Admin of a multisite. Haven't dug in any, but happy to if needed.

kraftbj:
`get_liveblog_state`, when in the admin, always assumes there is a global `$post` object set, which isn't the case quite often.

Without groking the overall internals of when this function is called, etc, returning out when it is not set fixes #220.

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -378,6 +378,9 @@ final class WPCOM_Liveblog {
 		}
 		if ( empty( $post_id ) ) {
 			global $post;
+			if ( ! $post ){
+				return false;
+			}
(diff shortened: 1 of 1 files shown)


grappler:
@philipjohn When do you plan to release an update so that I can profit from this fix?

philipjohn:
Soon soon, very soon!

Issue #224, opened 2015-12-10
Create a Liveblog key events widget


philipjohn:
We can't put the shortcode in a text widget, so let's create a widget for it. It'll need to have the options that correspond to the shortcode options, and then it can just do `do_shortcode()`

PR #226, opened 2016-01-13
Add widget to display a list of key events

rodrigoprimo:
I have created a first version of the widget to display a list of key events (issue #224). It is a simple wrapper around [liveblog_key_events] shortcode and its single parameter is the widget title. The number of key events shown is controlled by the Liveblog post configuration. This is a screenshot of the widget form:

Widget form

And here a screenshot of the widget displaying key entries when the user is viewing a Liveblog post:

Widget form

It would be great to have general comments or suggestions on this first version since I'm still familiarizing myself with this plugin. Besides that, I have two specific questions to ask:

* The issue description (#224) suggests the use of do_shortcode() inside the widget. I decided to call WPCOM_Liveblog_Entry_Key_Events::shortcode() directly as it is more straightforward and AFAIK there are no drawbacks. Am I missing something?

* I like unit tests and I'm happy to see that they are becoming more popular in the WP community in recent years. It was great to see that Liveblog has its own tests. But the problem is that I'm not sure if it is worth writing unit tests for something like a widget. On one hand, it is always great to have your code covered by tests. On the other hand, a widget basically outputs HTML and I tend to think that it is hard to setup and maintain tests for this kind of code. What do you guys usually do in situations like this?

### DIFF
--- a/classes/class-wpcom-liveblog-entry-key-events-widget.php
+++ b/classes/class-wpcom-liveblog-entry-key-events-widget.php
@@ -1,0 +1,103 @@
+<?php
+
+/**
+ * Class WPCOM_Liveblog_Entry_Key_Events_Widget
+ *
+ * Class to create a widget that displays a list of key events.
+ * This widget is just a wrapper for the shortcode
+ * [liveblog_key_events].
+ */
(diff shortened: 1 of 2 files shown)


rodrigoprimo:
I accidentally created this PR against master. I'm going to close it and open another one against develop.

PR #227, opened 2016-01-13
Add widget to display a list of key events

rodrigoprimo:
I have created a first version of the widget to display a list of key events (issue #224). It is a simple wrapper around [liveblog_key_events] shortcode and its single parameter is the widget title. The number of key events shown is controlled by the Liveblog post configuration. This is a screenshot of the widget form:

Widget form

And here a screenshot of the widget displaying key entries when the user is viewing a Liveblog post:

Widget form

It would be great to have general comments or suggestions on this first version since I'm still familiarizing myself with this plugin. Besides that, I have two specific questions to ask:

* The issue description (#224) suggests the use of do_shortcode() inside the widget. I decided to call WPCOM_Liveblog_Entry_Key_Events::shortcode() directly as it is more straightforward and AFAIK there are no drawbacks. Am I missing something?

* I like unit tests and I'm happy to see that they are becoming more popular in the WP community in recent years. It was great to see that Liveblog has its own tests. But the problem is that I'm not sure if it is worth writing unit tests for something like a widget. On one hand, it is always great to have your code covered by tests. On the other hand, a widget basically outputs HTML and I tend to think that it is hard to setup and maintain tests for this kind of code. What do you guys usually do in situations like this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ The command system has one inbuilt command:
 
 `/key`: Which defines an entry to a key event, it adds the meta key to entry `liveblog_key_entry`. A key event can be styled using the `.type-key` class.
 
-To display a key event box you can add the `[liveblog_key_events]` shortcode in your theme, e.g. in the sidebar. Entries used with the key command will be inserted to both this box and the main feed. It also acts an anchor system to jump to parts of the main feed. It's not necessary to include the shortcode for the /key command to be enabled.
+To display a key event box you can add the `[liveblog_key_events]` shortcode in your theme, e.g. in the sidebar, or you can use the Liveblog Key Events widget. Entries used with the key command will be inserted to both this box and the main feed. It also acts an anchor system to jump to parts of the main feed. It's not necessary to include the shortcode for the /key command to be enabled.
 
 An example of using the key command would be an author writing the following in the New Entry box:
 
--- a/classes/class-wpcom-liveblog-entry-key-events-widget.php
+++ b/classes/class-wpcom-liveblog-entry-key-events-widget.php
@@ -1,0 +1,103 @@
+<?php
+
+/**
+ * Class WPCOM_Liveblog_Entry_Key_Events_Widget
+ *
+ * Class to create a widget that displays a list of key events.
+ * This widget is just a wrapper for the shortcode
+ * [liveblog_key_events].
+ */
+class WPCOM_Liveblog_Entry_Key_Events_Widget extends WP_Widget {
+
+	/**
+	 * Called by WPCOM_Liveblog::load(), it attaches the
+	 * widget.
+	 */
+	public static function load() {
+		add_action( 'widgets_init', array( __CLASS__, 'widgets_init' ) );
+	}
+
+	/**
+	 * Registers the widget
+	 */
+	public static function widgets_init() {
+		register_widget( __CLASS__ );
+	}
+
+	/**
+	 * Configure widget
+	 */
+	public function __construct() {
+		$widget_ops = array(
+			'class_name' => 'liveblog-key-events-widget',
+			'description' => __( 'A list of key events displayed when the user is viewing a Liveblog post.', 'liveblog' ),
+		);
+
+		parent::__construct(
+			'liveblog-key-events-widget',
+			__( 'Liveblog Key Events Widget', 'liveblog' ),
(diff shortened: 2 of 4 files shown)


mdawaffe:
How does the widget know which post to show Liveblog Key Events for?

rodrigoprimo:
@mdawaffe, I kept the logic used by the shortcode. The shortcode displays key events of the current post. If the current post is not a Liveblog or if the user is not viewing the single of a post, nothing is displayed.

mdawaffe:
What are the use cases we want to support for this widget? Would any of them include needing to display the Key Events for a particular post no matter what else the rest of the site is showing?

rodrigoprimo:
I wonder why the Travis build failed. I'm unable to reproduce this error and it seems to be unrelated with the changes I made. I even configured my fork to use Travis and the build passed (https://travis-ci.org/rodrigoprimo/liveblog/builds/102459191). Maybe it was a temporary error and it is just a matter of running the build again? Could someone with permission please check if this solves the error? I have seen this happen in the WP-CLI repository a few times. Any other idea of what might be happening?

joshbetz:
@rodrigoprimo I think it was a fluke. I just re-ran the test and it passed.

rodrigoprimo:
@mdawaffe my initial idea was to support the single use case supported by the shortcode and preserve its behaviour.

Regarding your suggestion to display the key events for a particular post, it would be good to hear the opinion of the maintainers of the plugin and of those in direct contact with users of the plugin. @philipjohn can you help here or point us to who can (since you created the issue)?

On one hand, I think that it makes sense to add a new option to make the widget (and the shortcode) display key events from a given Liveblog post. On the other hand, it might be confusing to the users, since in one case the widget will appear for every page of the site (as widgets normally do) and in another case it will appear only for Liveblog posts.

@joshbetz thanks for re-running the tests.

philipjohn:
I can see a case where a news site running a Liveblog might want to display the key events on the home page, that would then link through to the Liveblog. So if we can build that in now, I think that would be a good and useful additional feature.

rodrigoprimo:
@philipjohn thanks for your input. Any comment on using WPCOM_Liveblog_Entry_Key_Events::shortcode() instead of do_shortcode() (see the description of the PR)?

philipjohn:
Have you had a look at what Liveblog does with the shortcode? Do we miss anything by not using `do_shortcode()`?

rodrigoprimo:
@philipjohn it seems to me that we don't miss anything by not using ```do_shortcode()```. All this function does is parse the shortcode string and call ```WPCOM_Liveblog_Entry_Key_Events::shortcode()```. There are no hooks inside ```do_shortcode()```.

I'm not sure what you meant by looking at what Liveblog does with the shortcode. Could you please be more specific?

rodrigoprimo:
> What are the use cases we want to support for this widget? Would any of them include needing to display the Key Events for a particular post no matter what else the rest of the site is showing?

My suggestion is that we focus this PR on adding the widget supporting the single use case that is supported by the shortcode. I will create another issue for the inclusion of another use case (displaying the Key Events for a particular post).

If that is ok, I think that this PR is ready to be reviewed and merged if there is nothing to be changed.

philipjohn:
> I'm not sure what you meant by looking at what Liveblog does with the shortcode. Could you please be more specific?

I was just curious to know if the plugin did anything before `WPCOM_Liveblog_Entry_Key_Events::shortcode()` but it looks like that's what `do_shortcode()` calls so what you suggest sounds like a fine approach.

rodrigoprimo:
> I was just curious to know if the plugin did anything before WPCOM_Liveblog_Entry_Key_Events::shortcode() but it looks like that's what do_shortcode() calls so what you suggest sounds like a fine approach.

Ok, thanks for your feedback.

rodrigoprimo:
I've created a separate issue for the idea of displaying in the widget the key events for a given post (#230) so that this PR can focus on the creation of the widget displaying key events only if in a single of a Liveblog post (the current behavior of the shortcode).

mdawaffe:
@philipjohn, I think this is ready to be merged. Do you agree, or do you think it's blocked by #230? I think #230 is a good improvement, but not a blocker.

philipjohn:
I'd prefer to see more escaping but yeah, this is good to go :)

PR #228, opened 2016-01-18
Add support for WebSocket

rodrigoprimo:
This PR is a working proof of concept of adding WebSocket support to the Liveblog plugin (issue #221). It uses socket.io, Redis and socket.io-php-emitter (responsible for sending messages from PHP to the socket.io server via Redis). I'm opening it to validate the softwares that I chose and the initial implementation before continuing. It is important to note that this is my first experience with Socket.io and Redis.

I chose to use Socket.io and Redis for this proof of concept because of Guillermo Rauch and Tony Kovanen talks about the subject on WordCamp San Francisco 2014 and WordCamp Europe 2015. I wasn't able to find any WordPress plugin that supported WebSocket to use as reference or to check for alternatives to Socket.io. I still intend to do a little more research to see if I find any worth checking. Please let me know if you think I should look at another WebSocket library.

In this PR the plugin will use WebSocket instead of AJAX polling to refresh the list of Liveblog entries in the connected clients whenever a entry is added, deleted or updated. The plugin is still using AJAX when the user create, delete or update a Liveblog entry since I wasn't able to find a compelling reason to switch that part to WebSocket.

## How to test

To test this PR it is necessary to have both Redis and Node.js installed and then, from the Liveblog plugin directory, run the following commands:

```
composer install     # to install socket.io-php-emitter
cd nodeapp
npm install     # to install socket.io and the adapter socket.io-redis
node app.js     # to start the Node app that starts the socket.io server
```

Finally you should add the constant ```LIVEBLOG_USE_SOCKETIO``` to wp-config.php and set it to true.

After following the steps above, if you open a Liveblog post in two different browser windows, it is possible to see that whenever a user changes the list of entries in one window, the list in the other window is refreshed in "real time".

## Next steps

After researching a bit more for alternatives to Socket.io and hearing feedback from others, if a decision is made to use the architecture proposed in this proof of concept, the next steps will be:

* Fix known issues
* Better organize the code created for this PR
* Write documentation

## Known issues

This is a list of known issues that will be fixed if a decision is made to proceed with the proposed architecture.

* Users without permission will see new or updated entries with both the edit and delete button.
* For now it is only possible to use Redis and Socket.io running on the same server as WordPress and in their default ports.


rodrigoprimo:
Closing, again I went to quick and created a PR against master instead of develop :disappointed:

PR #234, opened 2016-01-27
Move the lazyload loading to the template_redirect so that is_enabled…

### LINKED ISSUE #233: Lazyload fails to...... load
Somewhere a regression got in, stopping lazyload from ever doing it's thing.

It's triggered by `WPCOM_Liveblog_Lazyloader::late_load()` which is called on the `after_liveblog_init` hook. That hook is called on `init` at which point, there is no post info available.

So when `self::is_enabled()` gets called it always returns false.

H/t to @bcampeau for spotting and debugging this.

philipjohn:
…() doesn't just return false every time. Fixes #233


bcampeau:
:+1:

PR #237, opened 2016-02-25
Restore embedded videos that went missing after 1.5 upgrade

### LINKED ISSUE #235: Embedded videos are missing after upgrade to Liveblog 1.5
When user adds video e.g. from Youtube it is missing from the post.

Steps to reproduce:
1. Add link to youtube video.
2. Publish
3. lembedded video frame is missing.

Update 1.5 improved escaping by adding sanitation functions at output.
Post rendered by wp_embed is if filtered through wp_kses_post that cuts iframe tag.

@philipjohn  would it be a good idea to add iframe to allowedposttags?

/cc @nb, added for trial visibility.

budzanowski:
Moved `wp_kses_post` before `auto embed` and `do_shortcode` so it does not strip html code introduced by those functions.

Fixes #235

### DIFF
--- a/classes/class-wpcom-liveblog-entry.php
+++ b/classes/class-wpcom-liveblog-entry.php
@@ -110,15 +110,15 @@ class WPCOM_Liveblog_Entry {
 	public static function render_content( $content, $comment = false ) {
 		global $wp_embed;
 
+		$content = wp_kses_post( $content );
+
 		if ( apply_filters( 'liveblog_entry_enable_embeds', true ) ) {
 			if ( get_option( 'embed_autourls' ) )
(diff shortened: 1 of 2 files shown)


budzanowski:
I basically did two things.
I had to move the `wp_kses_post` from `liveblog-single-entry` template. This allowed me to move `wp_kses_post` before `autoembed` and `do_shortcodes`. This way we filter input from user and leave html generated by shortcodes and oembed. I know that the convention is to strip directly in the template but this seems like a special case (we are introducing HTML tags ourselves).

philipjohn:
See #235

PR #239, opened 2016-02-29
Add tags to allow embedded video playback. 

### LINKED ISSUE #235: Embedded videos are missing after upgrade to Liveblog 1.5

budzanowski:
As per discussion in #235 new expanded array of allowed html tags was created. It is used in template with `wp_kses` to filter post text. It is basically `wp_kses_post` + additional tags.
`iframe` was needed for embedded online content and `source` for local content.

Fixes #235

### DIFF
--- a/classes/class-wpcom-liveblog-entry.php
+++ b/classes/class-wpcom-liveblog-entry.php
@@ -15,6 +15,7 @@ class WPCOM_Liveblog_Entry {
 
 	private $comment;
 	private $type = 'new';
+	private static $allowed_tags_for_entry;
 
 	public function __construct( $comment ) {
 		$this->comment  = $comment;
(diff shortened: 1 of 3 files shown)


budzanowski:
PR updated with test to check sanitization of user input.
@philipjohn This PR now has all conclusions of #235 implemented.

PR #240, opened 2016-03-10
[WIP] Feature - Adding REST API endpoints

travisw:
Integration of new REST API framework 
Still a work in progress.

### DIFF
--- a/classes/class-wpcom-liveblog-entry-extend-feature-authors.php
+++ b/classes/class-wpcom-liveblog-entry-extend-feature-authors.php
@@ -76,7 +76,7 @@ class WPCOM_Liveblog_Entry_Extend_Feature_Authors extends WPCOM_Liveblog_Entry_E
 		$endpoint_url = admin_url( 'admin-ajax.php' ) .'?action=liveblog_authors';
 		$use_rest_api = 0;
 
-		if ( WPCOM_Liveblog::use_rest_api && WPCOM_Liveblog::can_use_rest_api() ) {
+		if ( WPCOM_Liveblog::use_rest_api() ) {
 			$endpoint_url = trailingslashit( trailingslashit( WPCOM_Liveblog_Rest_Api::build_endpoint_base() ) . 'authors');
 			$use_rest_api = 1;
 		}
--- a/classes/class-wpcom-liveblog-entry-extend-feature-authors.php
+++ b/classes/class-wpcom-liveblog-entry-extend-feature-authors.php
@@ -74,11 +74,9 @@ class WPCOM_Liveblog_Entry_Extend_Feature_Authors extends WPCOM_Liveblog_Entry_E
 	public function get_config( $config ) {
 
 		$endpoint_url = admin_url( 'admin-ajax.php' ) .'?action=liveblog_authors';
-		$use_rest_api = 0;
 
 		if ( WPCOM_Liveblog::use_rest_api() ) {
 			$endpoint_url = trailingslashit( trailingslashit( WPCOM_Liveblog_Rest_Api::build_endpoint_base() ) . 'authors');
-			$use_rest_api = 1;
 		}
 
 		// Add our config to the front end autocomplete
@@ -92,7 +90,6 @@ class WPCOM_Liveblog_Entry_Extend_Feature_Authors extends WPCOM_Liveblog_Entry_E
 			'regex'       => '@([\w\-]*)$',
 			'replacement' => '@${key}',
 			'template'    => '${avatar} ${name}',
-			'use_rest_api' => $use_rest_api,
 		) );
 
 		return $config;
--- a/classes/class-wpcom-liveblog-entry-extend-feature-authors.php
+++ b/classes/class-wpcom-liveblog-entry-extend-feature-authors.php
@@ -77,7 +77,7 @@ class WPCOM_Liveblog_Entry_Extend_Feature_Authors extends WPCOM_Liveblog_Entry_E
 		$use_rest_api = 0;
 
 		if ( WPCOM_Liveblog::use_rest_api && WPCOM_Liveblog::can_use_rest_api() ) {
-			$endpoint_url = trailingslashit( trailingslashit( trailingslashit( WPCOM_Liveblog_Rest_Api::build_endpoint_base() ) . WPCOM_Liveblog::$post_id ) . 'authors');
+			$endpoint_url = trailingslashit( trailingslashit( WPCOM_Liveblog_Rest_Api::build_endpoint_base() ) . 'authors');
 			$use_rest_api = 1;
 		}
 
--- a/classes/class-wpcom-liveblog-entry-extend-feature-authors.php
+++ b/classes/class-wpcom-liveblog-entry-extend-feature-authors.php
@@ -206,7 +206,7 @@ class WPCOM_Liveblog_Entry_Extend_Feature_Authors extends WPCOM_Liveblog_Entry_E
 		exit;
 	}
 
-	public function get_authors( $term = '' ) {
+	public function get_authors( $term ) {
 
 		// The args used in the get_users query.
 		$args = array(
--- a/classes/class-wpcom-liveblog-entry-extend-feature-authors.php
+++ b/classes/class-wpcom-liveblog-entry-extend-feature-authors.php
@@ -73,17 +73,26 @@ class WPCOM_Liveblog_Entry_Extend_Feature_Authors extends WPCOM_Liveblog_Entry_E
 	 */
 	public function get_config( $config ) {
 
+		$endpoint_url = admin_url( 'admin-ajax.php' ) .'?action=liveblog_authors';
+		$use_rest_api = 0;
+
+		if ( WPCOM_Liveblog::use_rest_api && WPCOM_Liveblog::can_use_rest_api() ) {
+			$endpoint_url = trailingslashit( trailingslashit( trailingslashit( WPCOM_Liveblog_Rest_Api::build_endpoint_base() ) . WPCOM_Liveblog::$post_id ) . 'authors');
+			$use_rest_api = 1;
+		}
+
 		// Add our config to the front end autocomplete
 		// config, after first allowing other plugins,
 		// themes, etc. to modify it as required
 		$config[] = apply_filters( 'liveblog_author_config', array(
 			'type'        => 'ajax',
 			'cache'       => 1000 * 60 * 30,
-			'url'         => admin_url( 'admin-ajax.php' ) .'?action=liveblog_authors',
+			'url'         => $endpoint_url,
 			'search'      => 'key',
 			'regex'       => '@([\w\-]*)$',
 			'replacement' => '@${key}',
 			'template'    => '${avatar} ${name}',
+			'use_rest_api' => $use_rest_api,
 		) );
 
 		return $config;
--- a/classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
+++ b/classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
@@ -79,7 +79,7 @@ class WPCOM_Liveblog_Entry_Extend_Feature_Hashtags extends WPCOM_Liveblog_Entry_
 		$use_rest_api = 0;
 
 		if ( WPCOM_Liveblog::use_rest_api && WPCOM_Liveblog::can_use_rest_api() ) {
-			$endpoint_url = trailingslashit( trailingslashit( trailingslashit( WPCOM_Liveblog_Rest_Api::build_endpoint_base() ) . WPCOM_Liveblog::$post_id ) . 'hashtags');
+			$endpoint_url = trailingslashit( trailingslashit( WPCOM_Liveblog_Rest_Api::build_endpoint_base() ) . 'hashtags');
 			$use_rest_api = 1;
 		}
 
--- a/classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
+++ b/classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
@@ -78,7 +78,7 @@ class WPCOM_Liveblog_Entry_Extend_Feature_Hashtags extends WPCOM_Liveblog_Entry_
 		$endpoint_url = admin_url( 'admin-ajax.php' ) .'?action=liveblog_terms';
 		$use_rest_api = 0;
 
-		if ( WPCOM_Liveblog::use_rest_api && WPCOM_Liveblog::can_use_rest_api() ) {
+		if ( WPCOM_Liveblog::use_rest_api() ) {
 			$endpoint_url = trailingslashit( trailingslashit( WPCOM_Liveblog_Rest_Api::build_endpoint_base() ) . 'hashtags');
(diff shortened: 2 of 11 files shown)


classes/class-wpcom-liveblog-entry-extend-feature-authors.php
```diff
@@ -197,6 +197,17 @@ public function add_author_class_to_entry( $classes, $class, $comment_id ) {
...
+		$term  = isset( $_GET['autocomplete'] ) ? $_GET['autocomplete'] : '';
+		$users = $this->get_authors( $term );
+
+		header( "Content-Type: application/json" );
+		echo json_encode( $users );
+
+		exit;
+	}
+
+	public function get_authors( $term = '' ) {
```
mjangda: Do we need a default value here?

classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
```diff
@@ -181,6 +181,24 @@ public function add_term_class_to_entry( $classes, $class, $comment_id ) {
...
+	}
+
+	/**
+	 * Get a list of hashtags matching the search term
+	 *
+	 * @param string $term The term to search for
+	 *
+	 * @return array Array of matching hastags
+	 */
+	public function get_hash_terms( $term ) {
```
mjangda: This might be better named `get_hashtag_terms`.

classes/class-wpcom-liveblog-rest-api.php
```diff
@@ -0,0 +1,496 @@
...
+ */
+
+class WPCOM_Liveblog_Rest_Api {
+
+	private static $api_version;
+	private static $api_namespace;
+
+	public static $endpoint_base;
+	public static $endpoint_get_entries_by_date;
+	public static $endpoint_crud;
```
mjangda: `endpoint_get_entries_by_date` and `endpoint_crud` look unused.

classes/class-wpcom-liveblog-rest-api.php
```diff
@@ -0,0 +1,496 @@
...
+	/**
+	 * Load everything the class needs
+	 */
+	public static function load() {
+
+		self::$api_version   = '1';
+		self::$api_namespace = 'liveblog/v' . self::$api_version;
+		if ( get_option( 'permalink_structure' ) ) {
+			// Pretty permalinks enabled 
+			self::$endpoint_base = '/wp-json/' . self::$api_namespace . '/';
```
mjangda: Can we get the API base programatically instead of hard-coding it? (I believe the `wp-json` part can change.)

classes/class-wpcom-liveblog-rest-api.php
```diff
@@ -0,0 +1,496 @@
...
+		WPCOM_Liveblog::$post_id          = $post_id;
+	}
+
+	/**
+	 * Validation callback to check for numeric value
+	 *
+	 * @return bool true if $param is numeric. false otherwise.
+	 */
+	public static function validate_is_numeric( $param, $request, $key ) {
+		return is_numeric( $param );
```
mjangda: `is_numeric( -1 ) === true` so this may not the best check for our params since most (all?) expect positive integers.

classes/class-wpcom-liveblog-rest-api.php
```diff
@@ -0,0 +1,496 @@
...
+	}
+
+	/**
+	 * Validation callback to check for allowed crud action
+	 *
+	 * @return bool true if $param is one of insert|update|delete|delete_key. false otherwise
+	 */
+	public static function validate_crud_action( $param, $request, $key ) {
+		return in_array( $param, array( 'insert', 'update', 'delete', 'delete_key' ) );
```
mjangda: This is duplicated in the `WPCOM_Liveblog` class and we should find a way to avoid the repetition.

liveblog.php
```diff
@@ -777,13 +882,20 @@ private static function add_default_plupload_settings() {
...
+			return $url;
+		} else {
+			$post_permalink = get_permalink( self::$post_id );
+			if ( false !== strpos( $post_permalink, '?p=' ) ) {
+				$url = add_query_arg( self::url_endpoint, '', $post_permalink ) . '='; // returns something like ?p=1&liveblog=
+			} else {
+				$url = trailingslashit( trailingslashit( $post_permalink ) . self::url_endpoint ); // returns something like /2012/01/01/post/liveblog/
+			}
+			$url = apply_filters( 'liveblog_endpoint_url', $url, self::$post_id );
+			return $url;
```
mjangda: Instead of repeating the filter and return twice, we can just move it outside the `if/else`.

liveblog.php
```diff
@@ -462,21 +498,39 @@ public static function ajax_crud_entry() {
...
+
+	/**
+	 * Perform a specific CRUD action on an entry for a post
+	 *
+	 * @param string $crud_action Allowed actions are insert|update|delete|delete_key
+	 * @param array $args An array of data to be passed to the crud method
+	 *
+	 * @return mixed The result of the crud method
+	 */
+	public static function do_crud_entry( $crud_action, $args ) {
```
mjangda: Should we move the `crud_action` validation here instead?

liveblog.php
```diff
@@ -1221,6 +1381,16 @@ private static function is_wp_too_old() {
...
 	}
+
+	/**
+	 * Checks to see if the current WordPress version has REST API support
+	 *
+	 * @return bool true if supported, false otherwise
+	 */
+	private static function can_use_rest_api() {
+		global $wp_version;
+		return version_compare( $wp_version, self::min_wp_rest_api_version, '>=' );
```
mjangda: Instead of a strict version compare, might be better to check if a REST API related class exists.

liveblog.php
```diff
@@ -34,13 +34,15 @@
...
-	const nonce_key        = 'liveblog_nonce';
+	const version                 = '1.4';
+	const rewrites_version        = 1;
+	const min_wp_version          = '3.5';
+	const min_wp_rest_api_version = '4.4';
+	const key                     = 'liveblog';
+	const url_endpoint            = 'liveblog';
+	const edit_cap                = 'publish_posts';
+	const nonce_key               = '_wpnonce'; // Using these strings since they're hard coded in the rest api. It'll still work fine for < 4.4
+	const nonce_action            = 'wp_rest';
```
mjangda: Just to clarify here: the REST API is expecting a nonce param called `_wpnonce` and the action always as `wp_rest`? 

The use of `_wpnonce` as the param could break in the admin context (e.g. when interacting with the metabox) so probably worth testing.

t/test-rest-api.php
```diff
@@ -0,0 +1,110 @@
...
+		curl_setopt( $ch2, CURLOPT_POST, 1 );
+		curl_setopt( $ch2, CURLOPT_POSTFIELDS, $post_data_insert );
+
+		curl_exec( $ch2 );
+		$response2_http_code = curl_getinfo( $ch2, CURLINFO_HTTP_CODE );
+		curl_close( $ch2 );
+
+		// HTTP response codes should be 403 Forbidden
+		$this->assertEquals( 403, $response1_http_code );
+		$this->assertEquals( 403, $response2_http_code );
```
mjangda: Should we split this up into smaller tests?

t/test-rest-api.php
```diff
@@ -0,0 +1,110 @@
...
+		
+	}
+
+	private static function set_common_curl_options( $_ch ) {
```
mjangda: Could we simplify/speed up the tests by calling methods directly and verifying the output instead of making curl calls?

mjangda: If we do end up sticking with remote calls, the WordPress HTTP API (https://codex.wordpress.org/HTTP_API) has lots of helper functions that can help simplify the code a lot.

classes/class-wpcom-liveblog-entry-extend-feature-authors.php
travisw: We shouldn't need that default value. Good catch.

classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
travisw: Makes sense. I'll change that.

classes/class-wpcom-liveblog-rest-api.php
travisw: Indeed they are unused, they shouldn't be there.

classes/class-wpcom-liveblog-rest-api.php
travisw: Ah, that's true. Everything that it's validating should also be `> -1`.

classes/class-wpcom-liveblog-rest-api.php
travisw: Repetition is bad, I'll fix this.

liveblog.php
travisw: Right, the REST API always uses those values.
I'll make sure to double check the admin context.

liveblog.php
travisw: Seems like a logical spot.

liveblog.php
travisw: Absolutely, good catch.

t/test-rest-api.php
travisw: Yeah, I do want to write some simpler, more focused tests. That's something I'll work on.
I'll take a look at the HTTP API too, it looks super helpful. Thanks!

liveblog.php
travisw: I ended up putting the `crud_action` validation in it's own method called `is_valid_crud_action`. This way we can still take advantage of the built in validation callback functionality in the REST API.

liveblog.php
travisw: I did originally consider going that route but then opted for the version check as it seemed to make the most sense. I'm curious what your thoughts are on the benefits of checking for the existence of a specific class.

classes/class-wpcom-liveblog-rest-api.php
travisw: Ah, yep, there's a function for that. It's been changed.

t/test-rest-api.php
```diff
@@ -319,6 +319,41 @@ function test_endpoint_get_entries() {
...
+	function test_endpoint_crud_action() {
+
+		// Create an author and set as the current user
+		$author_id = $this->factory->user->create( array(
+			'role' => 'author',
+		) );
+		wp_set_current_user( $author_id );
+
+		// Create a post
+		$this->factory->post->create();
```
mjangda: During the test we never actually convert the post into a liveblog; should the insert action pass?

t/test-rest-api.php
```diff
@@ -319,6 +319,41 @@ function test_endpoint_get_entries() {
...
+
+		// The POST data to insert
+		$post_vars  = $this->build_entry_args( array(
+			'crud_action' => 'insert',
+		));
+
+		// Try to access the endpoint and insert an entry
+		$request  = new WP_REST_Request( 'POST', self::ENDPOINT_BASE . '/1/crud/' );
```
mjangda: Will the post ID always be `1` here?

liveblog.php
```diff
@@ -734,7 +734,7 @@ public static function admin_enqueue_scripts( $hook_suffix ) {
 			$use_rest_api = 0;
 
 			if ( self::use_rest_api() ) {
-				$endpoint_url = trailingslashit( trailingslashit( trailingslashit( WPCOM_Liveblog_Rest_Api::build_endpoint_base() ) . $post->ID ) . 'post_state' );
+				$endpoint_url = WPCOM_Liveblog_Rest_Api::build_endpoint_base() . $post->ID . '/' . 'post_state';
```
mjangda: Thoughts on changing this to just be `state` (instead of `post_state`)? Or at least switching the `_` to `-`?

liveblog.php
```diff
@@ -897,7 +897,7 @@ private static function add_default_plupload_settings() {
 	 */
 	private static function get_entries_endpoint_url() {
 		if ( self::use_rest_api() ) {
-			$url = trailingslashit( trailingslashit( WPCOM_Liveblog_Rest_Api::build_endpoint_base() ) . self::$post_id );
+			$url = WPCOM_Liveblog_Rest_Api::build_endpoint_base() . self::$post_id . '/';
```
mjangda: Why do we have a trailing slash here but not on other endpoints?

t/class-wp-test-spy-rest-server.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+class WP_Test_Spy_REST_Server extends WP_REST_Server {
```
mjangda: We should credit the source of this class (I'm assuming it's from the WP-API test suite?)

t/test-rest-api.php
```diff
@@ -273,6 +294,31 @@ function test_update_post_state() {
...
+		$request = new WP_REST_Request( 'GET', self::ENDPOINT_BASE . '/1/entries/' . $start_time . '/' . $end_time . '/');
+		$response = $this->server->dispatch( $request );
+
+		// Assert successful response
+		$this->assertEquals( 200, $response->get_status() );
+
+		$entries = $response->get_data();
+
+		// The array should contain 1 entry
+		$this->assertCount( 1, $entries['entries'] );
```
mjangda: Could we do some deeper assertions here? (i.e. instead of just checking that we have one entry, that the entry data matches what we expect?)

travisw: We could indeed. My thinking was that these integration tests would be more high level. Checking to make sure that the endpoints exist and respond as expected. Then, checking to compare actual data in other more focused unit tests.
Though maybe this approach just shows my testing naivete, so any thoughts and suggestions are welcome.

liveblog.php
travisw: Yeah, that underscore doesn't make much sense with the rest of the URL structure.
I think just `state` would work fine here.

t/test-rest-api.php
travisw: Right, good catch. It does seem like the insert should fail if the post isn't a liveblog.
This is tied to the original functionality, I'd need to dig into that part of it.

t/test-rest-api.php
travisw: I'm making an assumption here that after the table is truncated and 1 record is inserted, that the ID of that record will be 1. But, that could be a dangerous assumption.
I have that ID of 1 hard coded in a few spots in the tests and that should be changed to grab the actual ID that was just inserted.

liveblog.php
travisw: When the javascript goes to use that `url` it expects it to already have the trailing slash, as you can see in `js/liveblog.js` (eg. `url += 'entries/‘;`).
This is just the base path and something else will always be appended in order to build a valid endpoint URL.

Whereas setting `$endpoint_url` in `admin_enqueue_scripts()` for example is building the full URL with nothing that needs to be appended, and so there’s no trailing slash in that case.

Are there other endpoints that you were referring to?

t/class-wp-test-spy-rest-server.php
travisw: Done

PR #247, opened 2016-06-13
Updated deduping logic for replaced entries to work properly with lazy loading

bcampeau:
Upon testing this, I noticed that liveblogs with edited entries were showing duplicates. It appears the logic in `remove_replaced_entries` was never updated when lazy loading was introduced. What's happening is that it's only searching the last X entries loaded for replaced entries, but not the entire result set. This leads to duplicates and (worse) old versions of entries becoming visible.

The only way to do this properly given the new logic is to not rely upon the `number` argument for `get_comments` and instead always fetch the full result set and handle deduping in PHP.

### DIFF
--- a/classes/class-wpcom-liveblog-entry-query.php
+++ b/classes/class-wpcom-liveblog-entry-query.php
@@ -40,7 +40,20 @@ class WPCOM_Liveblog_Entry_Query {
 	 * @param array $args the same args for the core `get_comments()`
 	 */
 	public function get_all( $args = array() ) {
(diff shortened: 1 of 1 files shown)


philipjohn:
Thanks @bcampeau - could you share some steps to reproduce so we can get this well tested please?

bcampeau:
Create a liveblog post without this fix with a number of entries greater than your pagination setting. Then update an entry that is on a page > 1. Refresh the post after editing and you'll see the new version of that entry pop up at the very top of the liveblog and also on whatever page where it previously existed.

PR #248, opened 2016-06-13
Auto load formatted tweets when new posts are either auto loaded at t…

bcampeau:
…he top of the timeline or lazy loaded when paginated

Otherwise, tweets load as plain text until the page is manually refreshed and lazy loaded tweets never get formatted.


philipjohn:
@bcampeau You are definitely one of this weeks favourite people, thank you!

bcampeau:
Glad I could be of service!

PR #256, opened 2016-09-02
Adding localisation to pluploads.js

david-binda:
Since the Liveblog's plupload implementation is not setting a custom container option, the default one (document.body) is being used.

This may introduce some issues for certain themes since the plupload container's css position is being set to relative.

This commit is adding `liveblog_plupload_localize` filter which can be used for changing the dropzone, browser and container options inside js/plupload.js


philipjohn:
Nicely done, thank you @david-binda

PR #266, opened 2016-09-13
Move is_robot function to javascript

### LINKED ISSUE #264: Using deprecated function `vary_cache_on_function`
Hi guys - I'm getting PHP notices thrown on every page load from liveblog pertaining to a deprecated function from `vip-deprecated.php` on a VIP Go local environment. 

`PHP Notice:  vary_cache_on_function is <strong>deprecated</strong> since version 2.0.0 with no alternative available. in /srv/www/mysite/wp-includes/functions.php on line 3709`

It looks like `classes/class-wpcom-liveblog-lazyloader.php` is attempting to call this function which is filling up our debug log.

Let me know if there's anything else that I can do to help track this bug down.

david-binda:
The Liveblog is currently reyling on an existence of `vary_cache_on_function` function when modifying the behaviour for bots.

As not all environments taking advantage of page level caching have the posibility of varying page cache on user agent, the check should happen in browser.

This commit it moving the logic to inside the js/liveblog-lazyloader.js taking advantage of the same regex and disabling the lazyloading initialisation for robots + hiding the load more button.

Fixes #264


philipjohn:
With the new React frontend I'm wondering whether we actually need this at all?

PR #271, opened 2016-09-30
Prevent old versions of edited entries from appearing in the feed.

philipjohn:
Fixes the logic in the updating function to;

* stop passing the ID when creating a new entry,
* pass the correct ID for making a comment as a replacement,
* stop updating the original comment with new content,
* mark the original comment as trash so only the new content is shown.

Also adds some inline commenting to describe the logic.


philipjohn:
While this PR addresses the issue, I'm not yet sure how it would affect Liveblogs with edited entries that pre-date this change.

Issue #273, opened 2016-10-17
Tests are broken since #270 was merged


david-binda:
As the `remove_replaced_entries` was renamed to `remove_replaced_deleted_entries` the unit tests are failing as they were not updated.

See https://github.com/Automattic/liveblog/pull/270/files and https://github.com/Automattic/liveblog/blob/e50c8dad22e7519f42a0045d5b564d3e60d9c5e4/t/test-entry-query.php#L61,L83

cc'ing @philipjohn

david-binda:
Seems like the failure due to changed method name was surpassed by following Fatal Error in PHP unit tests:

```
Fatal error: Call to a member function get_content() on a non-object in /home/travis/build/Automattic/liveblog/t/test-entry.php on line 57
```

That one is present since #271 

As the new logic introduced in #271 is going directly against the test case, the test might be obsolete and thus should be removed - see https://github.com/Automattic/liveblog/blob/e50c8dad22e7519f42a0045d5b564d3e60d9c5e4/t/test-entry.php#L53,L58

david-binda:
Fixing the Fatal error just for testing purposes revealed another failing test:

```
There was 1 failure:

1) Test_Entry::test_delete_should_return_entry_with_type_delete
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'delete'
+'new'
```

The change on line 26 in https://github.com/Automattic/liveblog/pull/271/files actually makes the [delete method of WPCOM_Liveblog_Entry](https://github.com/Automattic/liveblog/blob/master/classes/class-wpcom-liveblog-entry.php#L212,L234) not working as expected, as the replacement comment is returning type of "new" :-/

david-binda:
And there are also 2 more Warnings related to mentioned PRs ( #270  and #271 ):

```
1) Test_Entry_Query::test_remove_replaced_deleted_entries_should_remove_entries_replacing_other_entries
Undefined property: stdClass::$comment_approved

./liveblog/classes/class-wpcom-liveblog-entry.php:26
./liveblog/t/test-entry-query.php:64

2) Test_Entry_Query::test_remove_replaced_deleted_entries_should_not_remove_entries_replacing_non_existing_entries
Undefined property: stdClass::$comment_approved

./liveblog/classes/class-wpcom-liveblog-entry.php:26
./liveblog/t/test-entry-query.php:76
```

philipjohn:
[...]

PR #274, opened 2016-11-09
Better Cache Control

mjangda:
Improve caching for ajax requests:

- [x] Don't send nocache headers for lazyload + single entry reqs
- [x] Send a `Cache-Control: max-age={number}` header for entry fetch requests (should probably be filterable


mjangda:
Broken tests are unrelated (see #273)

PR #277, opened 2016-11-20
Initial implementation of AMP support

### LINKED ISSUE #246: Add support for AMP Live List
[AMP Live List](https://www.ampproject.org/docs/reference/extended/amp-live-list.html) could be an interesting way of supporting Liveblogs via AMP.

<strike>It would probably rely on having [WP-API support](https://github.com/Automattic/liveblog/issues/217) first, to give us the necessary endpoints.</strike>

jahvi:
Fixes #246

I started to work on AMP support for the liveblog plugin, my main goal is to try not edit template files unless absolutely necessary to keep them maintainable and not just add an extra AMP version of every single template, here's some input into my decisions:

- Added `is_amp` helper function to check AMP request, since I'm using it in a few places I figured it would be easier than checking with `function_exists...` everywhere.
- I hid the frontend liveblog editing capabilities using existing functions rather than removing the markup from the template.
- Made poll interval and items per page configurable with filters.
- For AMP live lists each entry needs a `data-sort-time` attribute with a timestamp, I noticed the single entry template has a `data-timestamp` attribute but it doesn't seem to be used anywhere, I renamed this to avoid duplication and having to mess with templates.
- Added simple styling borrowing from the current "Leave comment" button styles, might be worth making the update button fixed positioned so it's easily accesible.

Updating and deleting entries is a bit more complicated because of the way liveblog deals with these kind of entries:

- To support live updating entries we need to keep the same entry id and add a `data-update-time` attribute with the updated timestamp. I changed this so if a post is updated it'll take the timestamp of the original entry as the `data-sort-time` and the timestamp of the new entry as `data-update-time`, also the id attribute will always reference the original entry id.
- Deletions are trickier and I'm still looking into it, what I'm thinking is to change the `remove_replaced_deleted_entries` function so it doesn't unset deleted entries and display and empty `div` with the required `data-tombstone` attr and the entry id.

### DIFF
--- a/classes/class-wpcom-liveblog-entry-query.php
+++ b/classes/class-wpcom-liveblog-entry-query.php
@@ -46,14 +46,14 @@ class WPCOM_Liveblog_Entry_Query {
 		//
 		// We don't want to remove the parameter entirely for backwards compatibility
 		// since this is a public method, but we need instead to handle it as part
-		// of remove_replaced_entries after we retrieve the entire result set.
+		// of remove_replaced_deleted_entries after we retrieve the entire result set.
 		$number = 0;
 		if ( isset( $args['number'] ) ) {
 			$number = intval( $args['number'] );
 			unset( $args['number'] );
 		}
 
-		return self::remove_replaced_entries( $this->get( $args ), $number );
+		return self::remove_replaced_deleted_entries( $this->get( $args ), $number );
 	}
 
 	public function count( $args = array() ) {
@@ -102,7 +102,7 @@ class WPCOM_Liveblog_Entry_Query {
 			}
 		}
 
-		return self::remove_replaced_entries( $entries_between );
+		return self::remove_replaced_deleted_entries( $entries_between );
 	}
 
 	public function has_any() {
@@ -120,6 +120,15 @@ class WPCOM_Liveblog_Entry_Query {
 		return $all_entries_asc;
 	}
 
+	/**
+	 * Get deleted entries for a specific post.
+	 *
+	 * @return array The deleted entries IDs.
+	 */
+	public function get_deleted_entries() {
+		return ( array ) get_post_meta( $this->post_id, WPCOM_Liveblog_Entry::deleted_meta_key );
+	}
+
 	public static function entries_from_comments( $comments = array() ) {
 
 		if ( empty( $comments ) )
@@ -128,7 +137,7 @@ class WPCOM_Liveblog_Entry_Query {
 		return array_map( array( 'WPCOM_Liveblog_Entry', 'from_comment' ), $comments );
 	}
 
-	public static function remove_replaced_entries( $entries = array(), $number = 0 ) {
+	public static function remove_replaced_deleted_entries( $entries = array(), $number = 0 ) {
 		if ( empty( $entries ) ) {
 			return $entries;
 		}
@@ -136,9 +145,12 @@ class WPCOM_Liveblog_Entry_Query {
 		$entries_by_id = self::assoc_array_by_id( $entries );
 
 		foreach ( (array) $entries_by_id as $id => $entry ) {
+
+			// Removed replaced entries
 			if ( ! empty( $entry->replaces ) && isset( $entries_by_id[ $entry->replaces ] ) ) {
 				unset( $entries_by_id[ $id ] );
 			}
+
 		}
 
 		// If a number of entries is set and we have more than that amount of entries,
--- a/classes/class-wpcom-liveblog-entry.php
+++ b/classes/class-wpcom-liveblog-entry.php
@@ -172,9 +172,12 @@ class WPCOM_Liveblog_Entry {
 	 * @return WPCOM_Liveblog_Entry|WP_Error The newly inserted entry, which replaces the original
 	 */
 	public static function update( $args ) {
-		if ( !$args['entry_id'] ) {
-			return new WP_Error( 'entry-delete', __( 'Missing entry ID', 'liveblog' ) );
+		if ( ! $args['entry_id'] ) {
+			return new WP_Error( 'entry-update', __( 'Missing entry ID', 'liveblog' ) );
 		}
+
+		// Store the original entry.
+		$original_entry = $args;
 
 		// always use the original author for the update entry, otherwise until refresh
 		// users will see the user who editd the entry as  the author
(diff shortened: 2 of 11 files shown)


jahvi:
I just updated the PR with my initial implementation of the delete functionality, I had to make sure the element IDs didn't change so that the AMP page could pick up the the entry was deleted but I applied the same logic as the update functionality.

Right now it seems to work fine for new, updated and deleted entries, let me know if you have any feedback.

motion

philipjohn:
> For AMP live lists each entry needs a data-sort-time attribute with a timestamp, I noticed the single entry template has a data-timestamp attribute but it doesn't seem to be used anywhere, I renamed this to avoid duplication and having to mess with templates.

It looks like the `data-timestamp` is actually required for the lazyloading functionality. Take a look at [this code](https://github.com/Automattic/liveblog/blob/master/js/liveblog-lazyloader.js#L71) for context.

> To support live updating entries

Could you elaborate on what this means please? I think I'm perhaps just not familiar enough to understand, is this about having existing entries change in place?

> what I'm thinking is to change the remove_replaced_deleted_entries function

I may end up messing you about with this I'm afraid - that function is currently broken. I've made some changes in [fix/load-more-entries-button](https://github.com/Automattic/liveblog/tree/fix/load-more-entries-button) that are yet to be merged but which probably need to be considered. Tl;dr: the entry deletion is currently broken for lazyloading.

I'll have a look at the code now too and make any comments inline.

Oh one thing: can you help me with figuring out how to test this, as you have in your lovely GIF please?

classes/class-wpcom-liveblog-entry-query.php
```diff
@@ -142,11 +142,6 @@ public static function remove_replaced_deleted_entries( $entries = array(), $num
 				unset( $entries_by_id[ $id ] );
 			}
 
-			// Remove deleted entries
-			if ( 'delete' === $entry->get_type() ) {
-				unset( $entries_by_id[ $id ] );
-			}
-
```
philipjohn: I'm a bit worried about this change because this function is already kinda broken. See my changes in [fix/load-more-entires-button](https://github.com/Automattic/liveblog/commits/fix/load-more-entries-button)

classes/class-wpcom-liveblog-entry.php
```diff
@@ -23,7 +23,7 @@ public function __construct( $comment ) {
 		if ( $this->replaces && $this->get_content() ) {
 			$this->type = 'update';
 		}
-		if ( 'liveblog_trashed' === $comment->comment_approved ) {
+		if ( '' === $this->get_content() ) {
```
philipjohn: I'm worried this will show trashed entries when in looking in a normal browser with lazy loading.

templates/liveblog-loop-amp.php
```diff
@@ -0,0 +1,21 @@
+<?php if ( $show_archived_message ): ?>
+<div class="liveblog-archived-message">
+	<?php printf( __( '<strong>This liveblog is archived.</strong> If you need to publish new updates or edit or delete the old ones, you need to <a href="%s">enable it first</a>.' , 'liveblog'), get_edit_post_link() . '#liveblog' ); ?>
```
philipjohn: We should consider `get_edit_post_link()` to be untrusted and escape it's output.

jahvi:
> It looks like the data-timestamp is actually required for the lazyloading functionality. Take a look at this code for context.

My bad, I can add it back or edit the lazy loading script to use the new attribute name?

> Could you elaborate on what this means please? I think I'm perhaps just not familiar enough to understand, is this about having existing entries change in place?

Basically on top of loading new entries we also have to option to edit and delete entries, so if someone is watching a live blog on amp they can see entries edited and deleted in real time.

The way they do it is to add a `data-update-time` and `data-tombstone` attributes to entries that have been updated or deleted. The main problem I had was that because edits and deletions are handled as new entries the `id` attribute is changed to reflect this new entry instead of the original one and this needs to remain constant for the live list script to know what entries have been changed or deleted. I worked around this by keeping the `id` the same when you edit or delete entries.

> Tl;dr: the entry deletion is currently broken for lazyloading.

Thanks for letting me know I'll have a look at the branch and update the code accordingly.

> Oh one thing: can you help me with figuring out how to test this, as you have in your lovely GIF please?

What I did was just open a tab with a liveblog post and open a new tab with the AMP version of it (same url ending on /amp/) side by side. So when I made changes on the 'normal' view you can see them applied live on the amp version as well, there's a 15s minimum poll interval minimum on AMP live lists so changes are not real time.

templates/liveblog-loop-amp.php
jahvi: I'll make make the change to the normal template as well.

classes/class-wpcom-liveblog-entry-query.php
jahvi: I'll have a look at the branch I didn't know there was work being done on this, is it related to #108?

classes/class-wpcom-liveblog-entry.php
jahvi: This might not be needed if the functionality for deleting entries has been refactored, I'll have a look.

classes/class-wpcom-liveblog-entry-query.php
philipjohn: Looks like it is, although my reason for looking wasn't that issue :)

philipjohn:
> My bad, I can add it back or edit the lazy loading script to use the new attribute name?

I'd prefer we leave the original in place - folks might be using that and we wouldn't want to accidentally break their stuff.

jahvi:
Sounds good I added the `data-timestamp` attribute back and escaped the output of the `get_edit_post_link` function.

The new delete approach makes things significantly harder for the AMP live delete functionality though, as I can see from you're branch we're actually deleting entries instead of creating a new empty entry replacing the old one.

For the live AMP delete functionality to work we need to somehow keep track of the deleted entries so we can render an element like this `<div id="liveblog-entry-ID" data-tombstone></div>` that will let AMP know that that particular entry was deleted so the next time entries are polled it will automatically remove it from the DOM.

My initial idea would be to store the deleted entries IDs on the post meta, then render them somewhere in the post with the `data-tombstone` attribute, they don't have to be part of the main entries loop so it doesn't mess with the lazy loading functionality but still accessible by AMP.

I'll start playing around with this idea, let me know if you can suggest a more appropriate approach.

jahvi:
After merging the refactored delete entries functionality I also had to change the way I update entries on AMP pages.

Basically your new branch also introduces a change where edited entries 'inherit' the parents timestamp so they keep the same position on the thread. Before I could use the new entry's created time as the `data-update-time` attribute but now I keep track of it with a [comment meta on the new entry](https://github.com/Automattic/liveblog/pull/277/files#diff-d261b87cfa9200cd48f1e766e6497379R248).

I figured it would be a better idea to keep track of the original entry ID using comment metas as well, so now I'm doing that instead of updating the `liveblog_replaces` meta. I like how this works better because instead of changing functionality to accommodate AMP I'm extending what it's currently in place.

jahvi:
I updated the PR with refactored code to allow deleted entries to be removed live on AMP pages.

Because entries are actually deleted now I'm [keeping track](https://github.com/Automattic/liveblog/pull/277/files#diff-d261b87cfa9200cd48f1e766e6497379R282) of their IDs on the parent post, the on the AMP loop template I [render an empty `<div>`](https://github.com/Automattic/liveblog/pull/277/files#diff-9feb02e36f985f5a835554a54c1b5062R21) for each deleted entry with the `data-tombstone` attribute.

Everything seems to work fine now as far as I could test, including the lazy loading functionality.

jahvi:
Rebased on top of current master and squashed related commits.

jahvi:
I just pushed a commit to allow oEmbed elements, I though it was strange that they didn't work by default because the WP AMP plugin already [supports them](https://github.com/Automattic/amp-wp/tree/master/includes/embeds) out of the box.

It looks like there's a function hooked into the [comment_text](https://github.com/Automattic/liveblog/pull/277/commits/d15d6370ef716f893432c6e1ac600ec0497cc5b2#diff-d261b87cfa9200cd48f1e766e6497379R179) filter that's breaking the format of the oEmbed elements but I haven't been able to see exactly what it is. For now I'm just returning the unfiltered entry content on AMP pages but I'm still looking too see if there's a more elegant solution.

motion

jahvi:
After some debugging I found that the `force_balance_tags` filter was breaking the structure that AMP expects for oEmbed elements so instead of returning the unfiltered output I'm just [removing that filter instead](https://github.com/Automattic/liveblog/pull/277/commits/7dadd7fc89e7b2fdf7f91852e79d34ce63d37d83#diff-d261b87cfa9200cd48f1e766e6497379R178).

philipjohn:
@jahvi I realised we dropped this, sorry about that. I'd like to pick it back up again. Would you be up for bringing your branch in line with master as a first step?

jahvi:
No worries I'll get started on this then and update the PR

jahvi:
@philipjohn I just rebased master and everything seems fine on my end, let me know if it's ok.

GaryJones:
Hi @jahvi. I'm just going through old PRs. I can see that `master` branch has some support for AMP now built-in to it. Is there anything outstanding from your PR that could be added to it?

jahvi:
Hey @GaryJones I believe most of what I did has already been done on master so this can probably be ignored.

philipjohn:
Yep, we kinda leapfrogged on this, apologies. AMP support was added in #450 and release in v1.9

PR #278, opened 2016-11-20
Fix entries not getting removed after delete

jahvi:
Entries are not getting removed in real time after you click the delete button, this seems to be because the JS that does this is looking for an entry type of `delete` and deleted entries are getting the type of `new` which makes sense because that's what they actually are.

Basically [this part of the constructor](https://github.com/Automattic/liveblog/blob/master/classes/class-wpcom-liveblog-entry.php#L26-L28) is not running because the [new entry](https://github.com/Automattic/liveblog/blob/master/classes/class-wpcom-liveblog-entry.php#L232) won't have a `liveblog_trashed` comment approved value.

I'm assuming empty entries are deleted but that might not be ideal.

### DIFF
--- a/classes/class-wpcom-liveblog-entry.php
+++ b/classes/class-wpcom-liveblog-entry.php
@@ -23,7 +23,7 @@ class WPCOM_Liveblog_Entry {
 		if ( $this->replaces && $this->get_content() ) {
 			$this->type = 'update';
 		}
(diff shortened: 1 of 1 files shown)


jahvi:
Deleted entries are being refactored in (fix/load-more-entries-button)[https://github.com/Automattic/liveblog/commits/fix/load-more-entries-button] so this fix has been added as part of #277 in ...

PR #279, opened 2016-11-20
Fix WYSIWYG icons getting cropped in MAC OS

### LINKED ISSUE #275: WYSIWYG: Icons cut off in MAC OS
Some of the icons in the wysiwyg editor are being cut off in the latest versions of Chrome, Safari and Firefox on Mac OS (Sierra).

screen shot 2016-11-12 at 2 46 52 pm

I can submit a PR to address this if no one else is working on it at the moment.

jahvi:
Fixes #275

Increased wrapper element size to let full icon show.

| Before | After |
| --- | --- |
| before | after |

### DIFF
--- a/css/liveblog.css
+++ b/css/liveblog.css
@@ -149,9 +149,9 @@ a.liveblog-form-entry::-webkit-input-placeholder {
 	position: relative;
 	margin: 1px;
 	border: solid 1px rgba(0, 0, 0, 0.0);
-	padding: 2px 1px;
-	width: 20px;
-	height: 20px;
+	padding: 1px;
+	width: 24px;
+	height: 24px;
 	line-height: 20px;
 	cursor: pointer;
 }


css/liveblog.css
```diff
@@ -150,8 +150,8 @@ a.liveblog-form-entry::-webkit-input-placeholder {
 	margin: 1px;
 	border: solid 1px rgba(0, 0, 0, 0.0);
 	padding: 2px 1px;
-	width: 20px;
-	height: 20px;
+	width: 24px;
```
mjangda: I think the icons are supposed to squares. Is there a reason the width and height are different?

jahvi: It's because there's more vertical padding then horizontal, I basically tried to mimic `box-sizing: content-box;` to make sure the parent element would fit it's content instead of messing with background sizes or something like that. But I agree, I just reduced the padding as well so they can be made squares, I think they look better now.

screen shot 2016-11-21 at 09 31 10

PR #285, opened 2016-12-23
Contenteditable divs in Chrome

trepmal:
In Chrome (OS X and Windows 8 tested) and Safari, lines after the first are wrapped in a div. e.g.

if you enter

```
a
b
c
```

you get

```
a
<div>b</div>
<div>c</div>
```

This results in unexpected HTML when published. First line gets `p` tags, but the divs stick around.

While in some cases the issue can be worked around (#259, #265), it may be better to get to the root problem.

This pull request makes Chrome and Safari insert 2 breaks (`<br /><br />`) on enter which results in predictable spacing. Instead of the divs, you'd get

```
<p>a<br>
b<br>
c</p>
```

Which is the same as Firefox currently gets.

- linebreak fix: http://stackoverflow.com/questions/18552336/prevent-contenteditable-adding-div-on-enter-chrome
- browser detection: http://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser
- https://bugs.chromium.org/p/chromium/issues/detail?id=84936

_edit: spelling_

### DIFF
--- a/js/liveblog-publisher.js
+++ b/js/liveblog-publisher.js
@@ -161,6 +161,13 @@
 				},
 				'lineBreak': function() {
 					if ( isChrome || isSafari ) {
+						return e.shiftKey && e.keyCode === 13;
+					}
+
+					return;
+				},
+				'paragraphBreak': function() {
+					if ( isChrome || isSafari ) {
 						return e.keyCode === 13;
 					}
 
@@ -224,6 +231,10 @@
 			if (command === 'lineBreak') {
 				command = 'insertHTML';
 				value   = "<br><br>";
+			}
+			if (command === 'paragraphBreak') {
+				command = 'insertHTML';
+				value   = "<br><br><br>";
 			}
 			document.execCommand( command, false, value );
 		},
--- a/js/liveblog-publisher.js
+++ b/js/liveblog-publisher.js
@@ -4,9 +4,11 @@
 		return;
 	}
 
-	var isFirefox = typeof InstallTrigger !== 'undefined';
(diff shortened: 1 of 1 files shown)


trepmal:
Didn't realize my commit comments would get absorbed into the pull request.. sorry about the doubles :)

philipjohn:
Typical behaviour in the WordPress editor is that hitting **Return** will result in a new paragraph.

Would it be better have the Liveblog editor do the same for consistency? So instead of multiple `<br>` tags, we have each line wrapped in a `<p>` tag? Maybe putting the content through `wpautop()` would do the trick?

trepmal:
I was actually just thinking about that. For an experience more consistent with the standard post editor, we'd want `enter` to result in a new paragraph (I had trouble with wrapping the line, but an extra `<br>` would give the same effect :/), and `shift`+`enter` should give a single line break.

trepmal:
34af007 adds more WordPress-y behavior for line breaks. Hit enter, get a new paragraph. Hit shift+enter get a new line (same paragraph). I've gone a little cross-eyed and have not yet tested what happens in IE.

trepmal:
@philipjohn I think my pull request maybe got out of hand :) But any thoughts on the approach?

The first part, getting the editor in Chrome to handle like Firefox, is the main thing.

Getting the editor's line breaks to act like the WordPress post editor might be a nice enhancement, but could be put off till later if there's concern.

philipjohn:
Thanks! I've tested the latest state and it looks good - nice work :)

philipjohn:
Note failing tests are not related.

PR #292, opened 2017-05-29
Fixed twice updation issue at the same session

### LINKED ISSUE #243: Not possible to update entry twice at the same session.
**Steps to reproduce**
1. Create entry
2. Update that entry(do not refresh page afterwards)
3. Update that entry again.
4. Refresh page

**Outcome**
Second update is lost.

**Analysis**
After first update original entry is replaced with update entry. Second update tries to update update entry. When page is refreshed updates are filtered out, and we only see the original entry with text that was applied to this entry by the first update.

**Solution**
When preparing response to update ajax request prepare the request in such a way that it will have original entry id. This way second update will also target original entry.

emgk:
Fixes #243


rebeccahum:
@emgk This fix didn't seem to work. Perhaps it could be an ajax issue?

trepmal:
Hi @emgk,

Thanks for your contribution! I think we may have a simpler way forward, though. Would you mind checking if https://github.com/Automattic/liveblog/pull/297 works for you?

PR #293, opened 2017-05-30
Fixed showing deleted posts in 'load more' functionality

### LINKED ISSUE #244: Prefetched entries are not updated or deleted.
**Steps to reproduce**
1. Add several entries, as much as needed so on refresh they will not be all immediately visible. To show them you are required to press `LOAD MORE ENTRIES..` button.
2. Open second browser window and load post.
3. In the first window remove first entry that is not displayed in the second window.
4. Click `LOAD MORE ENTRIES..` in the second window

**Outcome**
On the second browser the entry that was just deleted is still visible. After page refresh it is not visible.

**Analysis**
Deleted entry was prefetched. Client gets information that there was delete but it only checks elements that are already displayed. Situation is the same for entry update.

Rather small gap but someone may complain.

**Solution**
Check if target entry is in lazyloader prefetched entries. If it is apply update, delete on it.

emgk:
Fixes #244

### DIFF
--- a/js/liveblog-lazyloader.js
+++ b/js/liveblog-lazyloader.js
@@ -99,6 +99,14 @@
 			} );
 		},
 
+		fetchAllEntries: function(){
+			var url = liveblog_settings.endpoint_url + 'lazyload/';
+
+			$.get( url, {}, function( response ) {
+			   console.log(response);
+			} );
+		},
+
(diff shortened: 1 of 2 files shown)


philipjohn:
Made redundant by the React front-end now

Issue #241, opened 2016-03-15
Remove Key Event from Key Events Widget does not work after entry update


budzanowski:
**Steps to reproduce:**
1. Enable Key Events Widget
2. Create entry with /key
3. Update that entry
4. Click X (delete) on Key Events widget for that entry.

**Effect:**
1. Key Event entry is removed from the Key Events Widget
2. Alert with `Error 500 Key Event Not Deleted` is displayed
3. After refresh Key Events is back in the Key Events Widget

**Analysis:**
After updating Key Event Entry in Liveblog this Key Events entry is replaced with updated version with different ID in blog and in Key Events Widget. Clicking delete tries to remove key event meta from the updated Event. Because of how the updates are processed the updated Entry underling `wp_comment` is missing key event meta. This is the underlying source of Alert because we try to delete non existing meta key.
After refresh it works again because how entries are picked for display. Updates are discarded and the original `wp_comments` are chosen for display. We can discard updates because in process of update original `wp_comment` was also changed. This propably should work differently and the updates should be chosen for display. The original `wp_comment` still has `key event meta` so it is again visible.

**Solution:**
To correct this error we should either rewrite which `wp_comments` are chosen for display. Ensure that Key Event update also has key event meta set and that the delete key event action removes `key event meta` in both update and original post.

budzanowski:
This should be related to #213.

maxhartshorn:
FYI we're still having issues with key event deletion. It seems sometimes when we delete a key event it is not deleted. I've replicated on our live site and on our local. My thought is that it's tied to this issue.

Any update on a fix?

philipjohn:
Fixed by #309

PR #297, opened 2017-06-06
users can edit more than 1x now and reloaded page will show

### LINKED ISSUE #177: Edited comments appear at end of liveblog after archiving
We noticed that if you edit a liveblog comment, it appears twice after the liveblog is archived. The edited comment appears at the bottom of the liveblog (which is now sorted in chrono order). The unedited comment appears in the same place as it originally appeared. According to VIP, 

> This is actually due to how the liveblog plugin "alters" posts and it's by design (oddly enough).
> When a post is edited, the original one is deleted and a new one is added, which has a more
> recent timestamp on it.
> 
> This is done in order to make sure browser caches are updated properly; if you check the GH
> code for Liveblog you'll see it pretty quickly.

One workaround solution is to keep the liveblog active. One would have to add a filter on liveblog_display_archive_query_args that sets the sort order to chrono. To make this elegant for the editors, it would be nice to have the ability to add extra checkboxes and knobs to the liveblog meta box area. I propose adding a do_action to liveblog/templates/meta-box.php wherein plugin developers can add their own metaboxes to control functionality of the liveblog:

do_action( 'liveblog_metabox_display', $state );

where $state is the current state of the liveblog (enabled, archived)

rebeccahum:
As per #243, users cannot edit an entry more than once on the same page session because it will only save once (ergo, the first edit).  Problem was in rendering html, since ajax was correctly.

Credits to @trepmal for really delving into it to figure the root cause.  

This also fixes #177 and #241 going forward.

### DIFF
--- a/classes/class-wpcom-liveblog-entry.php
+++ b/classes/class-wpcom-liveblog-entry.php
@@ -94,7 +94,7 @@ class WPCOM_Liveblog_Entry {
 	}
 
 	public function get_fields_for_render() {
-		$entry_id     = $this->comment->comment_ID;
+		$entry_id     = $this->replaces ? $this->replaces : $this->comment->comment_ID;
 		$post_id      = $this->comment->comment_post_ID;
 		$avatar_size  = apply_filters( 'liveblog_entry_avatar_size', self::default_avatar_size );
 		$comment_text = get_comment_text( $entry_id );


philipjohn:
@trepmal Would you be up for testing this?

PR #299, opened 2017-06-10
Entries that have been updated can now be "deleted"

rebeccahum:
As per #108, entries that were previously edited and then deleted were still being displayed after refresh.

### DIFF
--- a/classes/class-wpcom-liveblog-entry-query.php
+++ b/classes/class-wpcom-liveblog-entry-query.php
@@ -35,25 +35,6 @@ class WPCOM_Liveblog_Entry_Query {
 	}
 
 	/**
-	 * Query the database for all edited liveblog entries associated with $post_id
-	 *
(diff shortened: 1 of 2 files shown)


trepmal:
alternate approach: https://github.com/Automattic/liveblog/pull/159

#159 has some nice conversation on it about things to look out for. Might be good to compare the approaches and see what can be gleaned.

philipjohn:
What do you think @rebeccahum ?

rebeccahum:
@philipjohn both methods work great. Mine modifies the database so it appears correctly with the current code on how comments are rendered, whereas #159 changes the way comments that are "deleted" are rendered.

philipjohn:
Okay, I'm going to close this one in favour of #159. It'd be great to properly clear the entries but it looks like this will cause a frontend DB write which is potentially nasty, especially on a Liveblog.

PR #302, opened 2017-06-16
This addresses issue 288 by moving the index parameter to the URI path

### LINKED ISSUE #288: Remove reliance on GET params for single entry ajax requests
Scenario:

* A user loads a liveblog
* There are several key events shown in the key events widget
* Some of the key events have not loaded in the initial set of entries that are loaded by default
* The user clicks on one of the entries that isn't visible
* The plugin fires off an ajax request [appending an `index` GET param to the endpoint URL](https://github.com/Automattic/liveblog/blob/e50c8dad22e7519f42a0045d5b564d3e60d9c5e4/js/liveblog-lazyloader.js#L82-L85)
* The single entry is returned and displayed

See where the [ajax action is set up](https://github.com/Automattic/liveblog/blob/f82842828cad0c7b921459b10a6e534f9cbe5355/liveblog.php#L299) and the [request is processed](https://github.com/Automattic/liveblog/blob/f82842828cad0c7b921459b10a6e534f9cbe5355/liveblog.php#L519-L575).

We should refactor this to append an ID or some such to the endpoint URL, rather than a GET param, to avoid bypassing caching mechanisms like Batcache.

maxk42:
This should prevent caching issues.  Fixes #288

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -577,15 +577,15 @@ final class WPCOM_Liveblog {
 
 		// Get all Liveblog entries that are to be lazyloaded.
 		$entries = self::$entry_query->get_for_lazyloading(
-			isset( $fragments[1] ) ? (int) $fragments[1] : 0,
-			isset( $fragments[2] ) ? (int) $fragments[2] : 0
+			isset( $fragments[2] ) ? (int) $fragments[2] : 0,
+			isset( $fragments[3] ) ? (int) $fragments[3] : 0
 		);
 		if ( ! $entries ) {
 			do_action( 'liveblog_entry_request_empty' );
 
 			self::json_return( array(
 				'entries' => array(),
-				'index'   => (int) filter_input( INPUT_GET, 'index' ),
+				'index'   => (int) ( isset( $fragments[0] ) ? $fragments[0] : 0 ),
 			) );
 		}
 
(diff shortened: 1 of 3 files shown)


liveblog.php
```diff
@@ -577,15 +577,15 @@ public static function ajax_lazyload_entries() {
...
+			isset( $fragments[2] ) ? (int) $fragments[2] : 0,
+			isset( $fragments[3] ) ? (int) $fragments[3] : 0
 		);
 		if ( ! $entries ) {
 			do_action( 'liveblog_entry_request_empty' );
 
 			self::json_return( array(
 				'entries' => array(),
-				'index'   => (int) filter_input( INPUT_GET, 'index' ),
+				'index'   => (int) ( isset( $fragments[1] ) ? $fragments[1] : 0 ),
```
philipjohn: What happens if this gets set to `0`? How does the code deal with that eventually, and what is the effect for the user?

maxk42:
If it's set to zero it should return zero as expected.  That behavior
hasn't changed.  What's changing is if the input is not set, this sets the
index to zero instead of null, which is the intended default behavior if
I'm understanding this correctly.

philipjohn:
This is fixed by the new frontend in #322

PR #303, opened 2017-06-23
Adding an auto-archive feature.

### LINKED ISSUE #87: Auto-archive liveblogs after certain period of time
Most of the time editors don't have any good motivation to archive liveblogs. After the event they rarely visit the post and don't mind having the posting interface.

They don't really care about sparing CPU cycles and we can't blame them.

So instead of counting on them to archive the liveblogs, we should automatically archive them after a set period of time, let's say a week.

maxk42:
Liveblog-enabled posts older than the given limit (default: 7) days will become archived posts.

Can be disabled by the admin.

Fixes #87.

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -49,6 +49,7 @@ final class WPCOM_Liveblog {
 	private static $post_id               = null;
 	private static $entry_query           = null;
 	private static $custom_template_path  = null;
+	private static $auto_archive_days     = null;
 
 	/** Load Methods **********************************************************/
 
@@ -69,6 +70,9 @@ final class WPCOM_Liveblog {
 		self::add_admin_actions();
 		self::add_admin_filters();
 		self::register_embed_handlers();
+
+		// Globally set auto-archive option
+		self::$auto_archive_days = get_option( 'liveblog_auto_archive', 7);
 
 		WPCOM_Liveblog_Entry_Key_Events::load();
 		WPCOM_Liveblog_Entry_Key_Events_Widget::load();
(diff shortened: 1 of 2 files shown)


liveblog.php
```diff
@@ -70,6 +71,9 @@ public static function load() {
 		self::add_admin_filters();
 		self::register_embed_handlers();
 
+		// Globally set auto-archive option
+		self::$auto_archive_days = get_option( 'liveblog_auto_archive', 7);
```
philipjohn: It'd be good to make this filterable by plugins or themes.

liveblog.php
```diff
@@ -410,9 +416,25 @@ public static function get_liveblog_state( $post_id = null ) {
 		if ( 1 == $state ) {
 			$state = 'enable';
 		}
+		if ( 'enable' == $state && self::$auto_archive_days ) {
```
philipjohn: We should use a strict (`===`) comparison here to avoid unexpected results.

liveblog.php
```diff
@@ -410,9 +416,25 @@ public static function get_liveblog_state( $post_id = null ) {
...
+			$time_diff = 60 * 60 * 24 * $auto_archive_days;
+			if ( ( get_post_time( 'U', true, $post_id ) + $time_diff ) < current_time( 'U' ) ) {
```
philipjohn: Is this going to match the time the post was published? If so, it's probably better to match on the time of the last Liveblog entry, otherwise a Liveblog that runs for more than 7 days will get unexpectedly archived even though it's active.

liveblog.php
```diff
@@ -410,9 +416,25 @@ public static function get_liveblog_state( $post_id = null ) {
...
+		}
 		return $state;
 	}
 
+	/**
+	 * Convenience function to archive the current post
+	 * 
+	 * @param int $post_id
+	 */
+	public static function archive_post( $post_id ) {
```
philipjohn: It looks like this probably duplicates `set_liveblog_state()`

Issue #242, opened 2016-03-15
Dangerous combination of [liveblog_key_events] shortcode and /key command.


budzanowski:
**To reproduce:**
Use Liveblog with Key Events Widget.
Put the following text into Liveblog entry:
`[liveblog_key_events] /key`

**Outcome**
The blog post dies. It becomes unresponsive and it is not possible to refresh the page.

From the point of usability it does not make sense to make such entry but it is possible to do it.

philipjohn:
Yikes.

Probably need to add something like this into the shortcode handler:

```
if ( WPCOM_Liveblog::is_liveblog_post() ) {
    return;
}
```

philipjohn:
Actually, that's not the right conditional. We need an `is_entry()` helper of some sort

philipjohn:
Fixed by #309

PR #309, opened 2017-08-30
Milestones/1.6

ollywarren:
**Liveblog 1.6 Milestone**

Addresses the following Issues:

#243 Not Possible to update an entry twice in a session - Thanks to @rebeccahum (PR #297) 
#242 Dangerous Combination of live blog shortcode and key command - Courtesy of Big Bite Creative (This PR)
#241 Remove key events from Key Events Widget does not work after entry update - Thanks to @rebeccahum (PR #297)
#177 Edited Comments appear at the end of live blog after archiving - Thanks to @rebeccahum (PR #297 & #298 )

Adds the following Features:

#87 Auto-archive live blog after certain amount of time - Thanks to @maxk42 (PR #303) Tweaks by Big Bite Creative (This PR)


philipjohn:
> #243 Not Possible to update an entry twice in a session - Thanks to @rebeccahum (PR #297)

Verified fix.

> #242 Dangerous Combination of live blog shortcode and key command - Courtesy of Big Bite Creative (This PR)

Verified fix.

> #241 Remove key events from Key Events Widget does not work after entry update - Thanks to @rebeccahum (PR #297)

Verified fix.

> #177 Edited Comments appear at the end of live blog after archiving - Thanks to @rebeccahum (PR #297 & #298 )

Were you able to reproduce the issue on this one? I've just tried and can't reproduce the original issue to verify the fix. I've also asked @rebeccahum

> #87 Auto-archive live blog after certain amount of time - Thanks to @maxk42 (PR #303) Tweaks by Big Bite Creative (This PR)

Could we get the feedback from #303 incorporated for this? Also, I've just tested and it doesn't work for me. Here's what I did;

1. Created a new liveblog
2. Added a few entries
3. Changed the post date to 2017-08-24 and saved
4, Refreshed the liveblog

I expected to see the archive auto-archived, but it was still active.

I verified with `wp shell` that `get_post_time( 'U', true, 12 )` returned `1503583098` (or "2017-08-24 13:58:18").

ollywarren:
@philipjohn Ill look into these ASAP, I missed the #303 feedback so will sort this out and see if i can replicate your issues with adjusting the publish date.

With regards #177 I was able to re-create it locally, I used an older branch of the codebase and verified the behaviour, then merged the fix and tested both the command line fix for retrospective bugs and also verified it fixed the issue moving forward. For the sake of clarification Ill make a point of re testing though and verify its working.

ollywarren:
@philipjohn to clarify with regards the publish date, the logic doesn't deal with the publish date, so upon creating a new Liveblog post we store the current date + the configured offset (default 7 days) as a timestamp on the options table. Then we check if the current date has passed the "expiry date" and auto archive at that point. By setting the publish date back, you wont effect anything as the expiry date is still set to the original date + offset.

If an archived post, is "re-enabled" as in the user clicks on enable in the admin panel, the same applies. The current date is increased by the offset and that is stored as a timestamp, and when that date is passed the blog is auto-archived again.

Hope that makes sense.

Is there a use case for people moving the publish date of the post back in time? Happy to adjust as directed.

philipjohn:
> upon creating a new Liveblog post we store the current date + the configured offset (default 7 days) as a timestamp on the options table.

Ah I'd been looking at the code in #303 so hadn't spotted that. In any case, I think archiving should be dependent on activity. If someone runs a Liveblog for >7 days they're going to be annoyed when it suddenly archives itself mid-flow. In my review of #303 I suggested using the last entry date.

ollywarren:
@philipjohn no problems ill update the feature to use that date.

rebeccahum:
@philipjohn I was able to reproduce #177.

classes/class-wpcom-liveblog-entry.php
```diff
@@ -281,6 +291,32 @@ private static function user_object_from_comment_id( $comment_id ) {
...
+	 * @param array $args The new Live blog Entry.
+	 * @return mixed
+	 */
+	public static function handle_restricted_shortcodes( $args ) {
+
+		// Runs the restricted shortcode array through the filter to modify it where applicable before being applied.
+		self::$restricted_shortcodes = apply_filters( 'liveblog_entry_restrict_shortcodes', self::$restricted_shortcodes );
+
+		// Foreach lookup key, does it exist in the content.
+		foreach ( self::$restricted_shortcodes as $key => $value ) {
```
philipjohn: We should double check that `self::$restricted_shortcodes` is an array before using it, just in case.

classes/class-wpcom-liveblog-entry.php
```diff
@@ -281,6 +291,32 @@ private static function user_object_from_comment_id( $comment_id ) {
...
+
+			// Regex Pattern will match all shortcode formats.
+			$pattern = "/\[{$key}(.*?)?\](?:(.+?)?\[\/{$key}(.*?)?\])?/";
```
philipjohn: We should replace this with `get_shortcode_regex()` so that we're using whatever core uses to find shortcodes.

ollywarren:
@philipjohn Auto Archive now using the latest entry publish date to calculate auto archive date. Each new entry will extend the date by the pre-determined offset. Also removed the options page and implemented a filter to allow developer configuration of the archive days. Docs updated.

Normalized URLs to improve perfromance with caching.

Change for issue #310 - If user/window is not in focus we increase refresh interval to 30 seconds

Change for issue #311 - The refresh interval is passed in the response allowing it to be changed at any time. 

``` 
add_filter( 'liveblog_post_{$post_id}_refresh_interval', function( $refresh_interval ) {
    return 60; 
} ); 
```
```
add_filter( 'liveblog_refresh_interval', function( $refresh_interval ) {
    return 5; 
} );
```

classes/class-wpcom-liveblog-wp-cli.php
```diff
@@ -11,6 +13,105 @@ public function readme_for_github() {
 		$readme = $this->markdownify_headings( $readme );
 		echo $readme;
 	}
+	
+	/**
+	 * Fix wp_commentmeta table so archived liveblog posts comments display properly.
+	 *
+	 * @subcommand fix-archive
+	*/
+	public function fix_archive() {
```
philipjohn: This command is pretty quiet :) We should get some more output so that it's clear to the user what the command is doing. Progress might be useful, too.

philipjohn: Also, a dry-run option would be best.

js/liveblog-admin.js
```diff
@@ -9,17 +9,42 @@ jQuery( function( $ ) {
...
+		var data = {};
+
+		if (liveblog_admin_settings.use_rest_api == 1) {
+			var url = liveblog_admin_settings.endpoint_url;
+			data['state']                           = encodeURIComponent( $( this ).val() );
+			data['template_name']                   = encodeURIComponent( $( '#liveblog-key-template-name' ).val() );
+			data['template_format']                 = encodeURIComponent( $( '#liveblog-key-template-format' ).val() );
+			data['limit']                           = encodeURIComponent( $( '#liveblog-key-limit' ).val() );
+			data[liveblog_admin_settings.nonce_key] = liveblog_admin_settings.nonce;
+			var method = 'POST';
```
philipjohn: Will this mean all requests from the frontend are `POST` requests? As `POST`s will always be uncached, they'll create unnecessary load if that's the default.

liveblog.php
```diff
@@ -332,40 +359,74 @@ public static function ajax_entries_between() {
...
+	public static function get_entries_by_time( $start_timestamp, $end_timestamp ) {
+
+		// Set some defaults
+		$latest_timestamp  = null;
+		$entries_for_json  = array();
+
 		// Do not cache if it's too soon
 		if ( $end_timestamp > time() ) {
-			$response_args['cache'] = false;
+			//$response_args['cache'] = false; --> Potentially Depricated
```
philipjohn: Code that isn't used or needed should be removed completely, rather than commenting it out. If it's needed in the future it can be restored via the Git history.

liveblog.php
```diff
@@ -332,40 +359,74 @@ public static function ajax_entries_between() {
...
-		}
 
-		/**
-		 * Loop through each liveblog entry, set the most recent timestamp, and
-		 * put the JSON data for each entry into a neat little array.
-		 */
-		foreach( $entries as $entry ) {
-			$latest_timestamp   = max( $latest_timestamp, $entry->get_timestamp() );
-			$entries_for_json[] = $entry->for_json();
+		// if ( empty( $entries ) ) {
```
philipjohn: Code that isn't used or needed should be removed completely, rather than commenting it out. If it's needed in the future it can be restored via the Git history.

liveblog.php
```diff
@@ -332,40 +359,74 @@ public static function ajax_entries_between() {
...
-		do_action( 'liveblog_entry_request', $result_for_json );
+		if ( ! empty( $entries_for_json ) ) {
+			do_action( 'liveblog_entry_request', $result );
+		} else {
+			do_action( 'liveblog_entry_request_empty' );
+		}
 
-		self::json_return( $result_for_json, $response_args );
+		return $result;
+		// self::json_return( $result_for_json, $response_args );
```
philipjohn: Code that isn't used or needed should be removed completely, rather than commenting it out. If it's needed in the future it can be restored via the Git history.

liveblog.php
```diff
@@ -395,21 +456,44 @@ public static function is_viewing_liveblog_post() {
...
+			//Lets grab todays day, convert it to a timestamp and look for any set auto archive date.
+			$today = date('Y-m-d H:i:s' );
+			$today_timestamp = strtotime($today);
+			$expiry = get_post_meta( $post_id, self::$auto_archive_expiry_key);
+
+			//if we have an expiry date lets compare them and if the
+			// expiry is less than today i.e. its in the past lets archive the liveblog.
+			if( $expiry ) {
+				if( (int)$expiry[0] < $today_timestamp ) {
+					self::set_liveblog_state( $post_id, 'archive' );
```
philipjohn: Will this end up running on the front end?

liveblog.php
```diff
@@ -504,6 +588,33 @@ public static function ajax_crud_entry() {
 				'latest_timestamp' => null
 			), array( 'cache' => false ) );
 		}
+
+		// self::json_return( $entry );
```
philipjohn: Code that isn't used or needed should be removed completely, rather than commenting it out. If it's needed in the future it can be restored via the Git history.

liveblog.php
```diff
@@ -541,30 +672,34 @@ public static function ajax_single_entry() {
...
 
-		do_action( 'liveblog_entry_request', $result_for_json );
+		if ( ! empty( $entries_for_json ) ) {
+			// Entries found
+			$result_for_json['index']             = (int) filter_input( INPUT_GET, 'index' );
+			$result_for_json['nextTimestamp']     = $next_timestamp;
+			$result_for_json['previousTimestamp'] = $previous_timestamp;
 
-		self::json_return( $result_for_json );
+			//self::json_return( $result_for_json );
```
philipjohn: Code that isn't used or needed should be removed completely, rather than commenting it out. If it's needed in the future it can be restored via the Git history.

liveblog.php
```diff
@@ -576,52 +711,88 @@ public static function ajax_lazyload_entries() {
...
 		$entry_content = $entry_content['content'];
 		$entry_content = WPCOM_Liveblog_Entry::render_content( $entry_content );
 
 		do_action( 'liveblog_preview_entry', $entry_content );
 
-		self::json_return(
-			array( 'html' => $entry_content ),
-			array( 'cache' => false )
-		);
+		// self::json_return(
```
philipjohn: Code that isn't used or needed should be removed completely, rather than commenting it out. If it's needed in the future it can be restored via the Git history.

liveblog.php
```diff
@@ -1275,6 +1609,19 @@ private static function is_wp_too_old() {
 		}
 		return version_compare( $wp_version, self::min_wp_version, '<' );
 	}
+
+	/**
+	 * Returns refresh interval after filters have been run
+	 *
+	 * @return int
+	 */
+	public static function get_refresh_interval() {
```
philipjohn: Nice!

liveblog.php
ollywarren: I believe it runs on, all page loads and the original feature PR had it that way. I guess ideally this sort of function would be a CRON job / scheduled task that would run and check once a day or so and run any housekeeping required, is there a preferred way to handle this?

classes/class-wpcom-liveblog-wp-cli.php
ollywarren: @philipjohn all the other tweaks are completed Ill tackle this one tomorrow morning first thing .

ollywarren:
@philipjohn just to clarify the settings page, I'm not seeing it in my code base at all? I have removed it for 1.6 can you point me to where your seeing the Settings page?

The metabox on the actual post is still in place, but the settings page that controlled the Auto Archive feature has been removed.

js/liveblog-admin.js
philipjohn: Ignore this :)

liveblog.php
philipjohn: Possibly, for now I'm just concerned with whether this will cause front-end DB writes which would be bad bad bad.

classes/class-wpcom-liveblog-wp-cli.php
```diff
@@ -13,67 +13,96 @@ public function readme_for_github() {
...
+		WP_CLI::line( 'Found ' . $total_liveblogs . ' Live Blogs.' );
+
+		foreach ($posts->posts as $post) {
+
+			//Increment the count so we get a more human readable index, inital value becomes 1 rather than 0.
+			$current_liveblog ++;
+
+			//Tell the user what we are doing, but lets colour this one se we can see its a new Liveblog in the console output.
+			// Note: Assigned to a varibale to be echo'd out, calling WP_CLI::colorize directly didnt output.
+			$intro = WP_CLI::colorize( "%4 Processing Liveblog {$current_liveblog} of {$total_liveblogs} %n");
```
philipjohn: [WP_CLI::colorize()](https://make.wordpress.org/cli/handbook/internal-api/wp-cli-colorize/) only makes the string fancy, it doesn't output. This needs to be further wrapped by `WP_CLI::line()` rather than echoing.

classes/class-wpcom-liveblog-wp-cli.php
```diff
@@ -13,67 +13,96 @@ public function readme_for_github() {
...
+						continue;
+
+					} elseif( false == $is_dryrun ) {
+						// If this isnt a dry run we can run the database Update.
+						for ( $i = 0; $i <= count( $correct_ids ) - 1; $i++ ) {
+
+							// replace with correct meta_value
+							if ( $correct_ids[$i] < $entry_id ) {
+
+								$wpdb->update(
```
philipjohn: Let's move the dry run check from above to here, so we can report in the dry run on what IDs are changing.

classes/class-wpcom-liveblog-wp-cli.php
```diff
@@ -85,16 +114,16 @@ public function fix_archive() {
...
+					INNER JOIN $wpdb->comments
 					ON $wpdb->comments.comment_id = $wpdb->commentmeta.comment_id
 					WHERE comment_post_id = %d
 					ORDER BY meta_value ASC", $post_id )
 				);
 
-			// check to make sure entry content being replaced matches available
-			if ( count( $entries_replace ) === count( $correct_contents) ) {
+			// check to make sure entry content being replaced matches available and that this is not a dry run
+			if ( count( $entries_replace ) === count( $correct_contents) && false === $is_dryrun ) {
```
philipjohn: Let's move this dry run check until just before the update action, so we can better report on what will be done in live mode.

philipjohn:
> just to clarify the settings page, I'm not seeing it in my code base at all? I have removed it for 1.6 can you point me to where your seeing the Settings page?

Ah, it's just the [template](https://github.com/Automattic/liveblog/pull/309/files#diff-1c6c4f85377a6b306bcfbc51ab1189c2R4)

classes/class-wpcom-liveblog-wp-cli.php
ollywarren: 👍

js/liveblog-admin.js
ollywarren: 👍

liveblog.php
ollywarren: @philipjohn does this require any action?

ollywarren:
@philipjohn I have now completed the required changes pending your review. Ive stripped out that template file for the settings page as its no longer required and I have moved the Auto Archive expiry check to a CRON job that runs every day at midnight. It was doing a single UPDATE request on the front end so this issues has been addressed and eliminated.

Let me know if theres anything else you need.

classes/class-wpcom-liveblog-cron.php
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public static function execute_auto_archive_housekeeping() {
+
+        //If Auto Archive is enabled,
+        if ( null !== WPCOM_Liveblog::$auto_archive_days ) {
+
+            $posts = new WP_Query( array(
+                'order'      => 'ASC',
+                'orderby'    => 'ID',
+                'meta_key'   => 'liveblog',
```
philipjohn: Could we use `WPCOM_Liveblog::$auto_archive_expiry_key` here instead?
Also, how many posts will this query return - will it be enough to catch all the Liveblogs we'll need to archive?

classes/class-wpcom-liveblog-cron.php
```diff
@@ -0,0 +1,68 @@
...
+                $post_id = $post->ID;
+
+                //Lets grab todays day, convert it to a timestamp and look for any set auto archive date.
+                $today = strtotime( date('Y-m-d H:i:s' ) );
+                $expiry = get_post_meta( $post_id, WPCOM_Liveblog::$auto_archive_expiry_key );
+
+                //if we have an expiry date lets compare them and if the
+                // expiry is less than today i.e. its in the past lets archive the liveblog.
+                if ( $expiry ) {
+                    if ( (int)$expiry[0] < $today ) {
```
philipjohn: Rather than use `$index[0]` could we just set the third parameter on `get_post_meta()` to `true`?

classes/class-wpcom-liveblog-cron.php
```diff
@@ -0,0 +1,68 @@
...
+                        WPCOM_Liveblog::set_liveblog_state( $post_id, 'archive' );
+                        $state = get_post_meta( $post_id, WPCOM_Liveblog::key, true );
```
philipjohn: Is this needed?

classes/class-wpcom-liveblog-cron.php
ollywarren: Implemented, makes sense, much simpler.

classes/class-wpcom-liveblog-cron.php
ollywarren: Ive taken this out.

classes/class-wpcom-liveblog-cron.php
ollywarren: @philipjohn ```WPCOM_Liveblog::$auto_archive_expiry_key``` references the key on the meta table that holds the expiry date, in the query above I'm looking for ALL posts that have a liveblog metakey indicating its been activated as a Liveblog post. Then I loop them check any set expiry date and update as required.

If preferred I could just grab only those with a meta_value of 'enabled'? so we skip anything thats already archived.

PR #324, opened 2017-12-20
Add filter for the_content only once to prevent issues with incorrect…

PavelK27:
… placement of new entry form.


philipjohn:
Thanks for this! Could you share with us the issue this PR attempts to resolve and the exact steps to reproduce please? Thanks!

PR #328, opened 2018-01-08
Author Select

liam-defty:
**To be merged after milestone/1.7**

This pull request contains the ability to be able to change the author and add contributors to an entry. As this is a feature that may not be needed by everyone you can activate it by defining a global variable in your `wp-config.php`:

`define('LIVEBLOG_ALLOW_AUTHOR_SELECT', true);`

When this feature is enabled the editor now has two drop downs below it allowing you to select and edit the authors and contributors. The author updates the actual author of the comment and the contributors are stored as comment meta as an array of ids.


classes/class-wpcom-liveblog-entry.php
```diff
@@ -11,7 +11,13 @@ class WPCOM_Liveblog_Entry {
 	 * @var string In case the current entry is an edit (replaces) of
 	 * another entry, we store the other entry's ID in this meta key.
 	 */
-	const replaces_meta_key   = 'liveblog_replaces';
+	const replaces_meta_key   	= 'liveblog_replaces';
```
philipjohn: Minor, but can we fix the spacing here?

liveblog.php
```diff
@@ -275,6 +275,24 @@ public static function get_post_id() {
 		return self::$post_id;
 	}
 
+	/**
+	 * Get current user
+	 */
+	public static function get_current_user() {
+		if (!self::is_liveblog_editable()) {
```
philipjohn: Could do with updating the spacing here.

liveblog.php
```diff
@@ -275,6 +275,24 @@ public static function get_post_id() {
...
+			return false;
+		}
+
+		$user = wp_get_current_user();
+
+		return array(
+			'id' => $user->ID,
+			'key' => strtolower($user->user_nicename),
```
philipjohn: Spacing :)

src/react/containers/EditorContainer.js
```diff
@@ -74,6 +95,29 @@ class EditorContainer extends Component {
...
+  onSelectAuthorChange(value) {
+    this.setState({
+      selectedAuthor: value,
+    });
+  }
+
+  getUsers(text, callback) {
+    const { config } = this.props;
+    getAuthors(text, config)
+      .timeout(10000)
```
philipjohn: Is this 10,000 seconds? Seems a bit much.

liam-defty: Milliseconds, so its 10 seconds :)

liam-defty:
@philipjohn That sounds like a good idea 👍  We'll see what we can come up with.

src/react/containers/EditorContainer.js
philipjohn: *wipes brow*

philipjohn:
I feel like the UI could be prettier, but it does the job. If we end up with a need for more options in future, we can move to a more discrete UI element.

One issue - when selecting a user as both the author and a contributor, they show up twice in the entry.

fireshot capture 4 - liveblog wordpress dev - https___localhost_2018_01_24_liveblog_

PR #329, opened 2018-01-09
Sharable Entries

liam-defty:
**to be merged after milestone/1.7**

A sharable link will be a link the the post with a hash of the entry id. When a user visits a url with a hash we will check if the entry id exists and jump to it on the front end using a similar approach and API to the jump to key events.

As it is more up to the theme developer to decide what social links they would like we do not include any on the actual entries by default. However here is a simple example of creating a share link to twitter:

```
add_filter( 'comment_text', function( $content, $comment ) {
	if ( $comment->comment_type === 'liveblog' ) {
		$share_link = get_permalink( $comment->comment_post_ID ).'#'.$comment->comment_ID;
		return $content.'<a href="https://twitter.com/intent/tweet?url='.urlencode( $share_link ).'">Share to Twitter</a>';
	} 

	return $content;
}, 10, 2 ); 
```


liam-defty:
@philipjohn This makes sense to do. Our designer, Ollie, suggested that we could follow a similar format to twitter where the date/time is a link to the entry, meaning you could copy the url from here. 

image

What are your thoughts?

philipjohn:
Just a note that the failing tests are fixed in #344

PR #334, opened 2018-01-15
Fix #320: Always return a full URL when building endpoint base

### LINKED ISSUE #320: RestAPI requests failing on multisites
I've got a subdirectory multisite that's returning returning `false` when attempting to enable liveblogging for a post. 

I've tracked it down some, but still haven't definitively narrowed it down to why. My theory is this; that the `get_post( $post_id )` call within the `admin_set_liveblog_state_for_post()` method is returning false because `WP_Post::get_instance()` (which uses the global var `$wpdb`) is looking in the wrong posts table. So for some reason the `get_post()` that's run from the API POST request isn't properly accounting for the `blog_id`. Instead it's just assuming `*_posts`. And if no post with the ID exists within the `_posts` table it returns false.

I've been able to semi-prove this theory by placing a `switch_to_blog()` call just above the `get_posts()`. 

So giving the above; I have a couple questions. 

- What am I missing? Is this newest version (1.6) still compatible with multisites as it's been in the past?
- Or what about adding a filter within `use_rest_api()` so that ` $use_rest_api` can be set to false so the JS will use the old ajax (`admin_ajax_set_liveblog_state_for_post`) function?

justnorris:
Fixes self-hosted multisite issue #320

### DIFF
--- a/classes/class-wpcom-liveblog-rest-api.php
+++ b/classes/class-wpcom-liveblog-rest-api.php
@@ -46,7 +46,7 @@ class WPCOM_Liveblog_Rest_Api {
 			$base = '/?rest_route=/' . self::$api_namespace . '/';
 		}
 
-		return $base;
+		return home_url( $base );
 
 	}
 
--- a/t/test-rest-api.php
+++ b/t/test-rest-api.php
@@ -78,7 +78,7 @@ class Test_REST_API extends WP_UnitTestCase {
 		$this->assertNotNull($base);
 
 		//Now assert the return matches the expected return.
-		$expected = '/?rest_route=/' . $api_namespace . '/';
+		$expected = home_url('/?rest_route=/' . $api_namespace . '/');
 
 		$this->assertSame($expected, $base);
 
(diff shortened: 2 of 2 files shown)


justnorris:
REST requests were sending requests to a path`/wp-json/liveblog/v1/6/get-entries/1/null-null` that was automatically resolved to be something like this: `http://localhost/wp-json/liveblog/v1/6/get-entries/1/null-null`

In a multisite environment where a site slug is set to be `news-site` the proper URL to request would be `http://localhost/news-site/wp-json/liveblog/v1/6/get-entries/1/null-null`. Because a relative URL was returned, the sub-site was never appended to the URL and was never queried.

The problem wasn't present in the wpcom environments because WordPress.com has special filters set up and have a custom endpoint URL: https://github.com/Automattic/liveblog/blob/master/wpcom-helper.php#L23

However, because the tests also used relative paths - the tests failed, so I changed the tests as well.

sboisvert:
I absolutely love the 2nd comment. Really well detailed. 💯

PR #335, opened 2018-01-15
Update plugin PHP version requirements

justnorris:


philipjohn:
Thanks @justnorris ! Could you run through how you decided to set it at 5.6? Was that based on a look at what PHP versions the plugin is compatible with? If so, what were the results? Could the plugin support lower versions relatively easily to get closer to parity with core?

justnorris:
Sure thing,
Core, at least currently, doesn't act on the "Requires PHP" tag, it just displays it and I thought it might be a good idea to match the version that [Travis CI checks for 5.6](https://github.com/Automattic/liveblog/blob/master/.travis.yml#L13-L23) is running

[Socketio class](https://github.com/Automattic/liveblog/blob/e50c8dad22e7519f42a0045d5b564d3e60d9c5e4/classes/class-wpcom-liveblog-socketio-loader.php#L18) requires at least PHP 5.3, so that would be the minimum, but it looks like it fails gracefully. 

I tested PHP 5.2.17 and it also works. If you want to ensure core parity, the requires version should probably be changed to 5.2.4, but then Travis CI should probably also run at version.

philipjohn:
Great, thanks for taking a look at that, and that seems like a sound decision process. Given 5.3 seems fine and that's the minimum for the socket stuff, let's go with 5.3 for now, and we can look at bumping it up later.

GaryJones:
The `readme.txt` [now](https://github.com/Automattic/liveblog/commit/01e72a474d46097833b97302eb566e0357c4c64a) has PHP 5.6 as the minimum, so this PR can be closed.

Issue #314, opened 2017-10-01
Please add schema LiveBlogPosting


siamlottery:
Hello, 
Please add Microdata to the liveblog plugins?
http://schema.org/LiveBlogPosting
Thank you very much.

philipjohn:
Nice suggestion, thanks!

justnorris:
I was thinking of taking this on today, but I noticed that Liveblog entries are rendered by React which complicates things a bit for inlining schema. Would it be better to generate an JSOND instead with a custom loop instead?

/cc @jasonagnew @liam-defty

jasonagnew:
Hey @justnorris 

That would be great. From my understanding, this helps search engines or other services better understand the page, therefore, it doesn't need updating on once the page loads. The Verge has an example of it: https://live.theverge.com/apple-live-blog-2017-iphone-8-event/ (Line 199)

So this could be outputted in the wp_head when the page loads using PHP / WordPress filter.

Does that sound correct to you?

Thanks
Jason

justnorris:
Yes, that's what I was thinking - but it's going to add a query to the initial page load, just making sure that's okay.

jasonagnew:
Awesome, the `get_all_entries_asc()` does get cached so I don't see an issue with this. It's also worth noting that before 1.7 the first batch of entries was added using PHP so that query was originally being called on page load. We only took it out in 1.7 to avoid having some entries coming from page load and others from the API, it made more sense to have everything come from the API.

justnorris:
Perfect, I'm going to work on this soon then :)

PR #337, opened 2018-01-16
Add LiveBlogPosting Schema

justnorris:
Append a LiveBlogPosting schema to a post that has a Liveblog enabled. 

Fills feature request #314


classes/class-wpcom-liveblog-schema.php
```diff
@@ -0,0 +1,155 @@
...
+			return $content;
+		}
+
+		try {
+			$post_id = WPCOM_Liveblog::get_post_id();
+		} catch ( Exception $e ) {
+			$post_id = get_the_ID();
+		}
+
+		$schema = new static( $post_id );
```
sboisvert: The use of static as a reference to itself was interesting. I had never seen it in use and was at first a bit confused. I looked it up and it's great and will work but I'm not sure what we gain vs the more traditional new WPCOM_Liveblog_Schema() or new self(); for example. Perhaps it's just me who hasn't had any experience with it but I think a comment here could be useful.

sboisvert:
I'm going to CC @david-binda to see if he's interested in giving code feedback since his feedback is always great.

nikkhilsharma:
Hey Guys,
Any update?

liveblog.php
```diff
@@ -305,7 +306,13 @@ public static function handle_request() {
 			// we need to add the liveblog after the shortcodes are run, because we already
 			// process shortcodes in the comment contents and if we left any (like in the original content)
 			// we wouldn't like them to be processed
+
+			// Add Liveblog schema right before the content
+			add_filter( 'the_content', array( 'WPCOM_Liveblog_Schema', 'add_schema_to_content' ), 20);
```
david-binda: I'm really not sure that `the_content` is the best place for performing the schema generation. `the_content` can be called multiple times during a page load, eg.: during secondary loops. Shall we perhaps either put the schema to head of the document, or call the `WPCOM_Liveblog_Schema ::add_schema_to_content` directly from within the `WPCOM_Liveblog::add_liveblog_to_content` method?

classes/class-wpcom-liveblog-schema.php
justnorris: `new static()` does the same thing as self, but is going to allow for extending the class much better. It's been around in PHP for quite some time, I think it's pretty safe to use. I can add a comment to clarify.

classes/class-wpcom-liveblog-schema.php
```diff
@@ -0,0 +1,155 @@
...
+			"name"              => get_the_title( $this->post_id ),
+			"coverageStartTime" => $start_time,
+			"liveBlogUpdate"    => $updates,
+		);
+
+		if ( ! empty( $end_time ) ) {
+			$schema['coverageEndTime'] = $end_time;
+		}
+
+		return apply_filters( 'liveblog_schema_liveblogposting', $schema );
```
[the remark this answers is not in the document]
justnorris: Adding an id or even access to `$this` would be really useful. The use case would be with, for example, custom content types. There are many items in the schema I didn't include that someone may want to add, which is why I added the filter there. But yeah - `$this` makes sense.

classes/class-wpcom-liveblog-schema.php
```diff
@@ -0,0 +1,155 @@
...
+	 * Get the coverage end time, based on the last entry
+	 *
+	 * @param $entries_asc
+	 *
+	 * @return string
+	 */
+	public function get_end_time( $entries_asc ) {
+
+		if ( 'archive' === WPCOM_Liveblog::get_liveblog_state() && is_array( $entries_asc ) ) {
+			$latest_update = $entries_asc[ count( $entries_asc ) - 1 ];
```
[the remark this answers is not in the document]
justnorris: Good catch, thanks.

liveblog.php
justnorris: At the moment I'm adding the schema exactly where the Liveblog is being added as well. I didn't want to re-do all the logic inside `handle_request`. Adding it as a filter to the_content provided an option to remove it, if I nest it inside `add_liveblog_to-content` it's going to be sticked. That said, - it totally makes sense for the the schema to be printed directly from `add_schema_to_content` - I'll change that and just add a filter (in case someone doesn't want a schema or they have already built their own schema).

classes/class-wpcom-liveblog-schema.php
```diff
@@ -108,11 +107,18 @@ public function get_live_blog_updates() {
...
 	public function get_start_time( $entries_asc ) {
-		if ( ! isset( $entries_asc[0] ) || ! isset( $entries_asc[0]['datePublished'] ) ) {
+
+		if ( ! is_array( $entries_asc ) || empty( $entries_asc ) ) {
+			return '';
+		}
+
+		$first_value = array_shift( $entries_asc );
+
+		if ( ! empty( $first_value['datePublished'] ) ) {
```
david-binda: I guess that this should had been `if ( empty( $first_value['datePublished'] ) ) {` (`!` is extra)

justnorris: 😨

david-binda: ehm, unit tests, ehm 😉

justnorris:
Done & Done.
This is ready for merging.

@philipjohn Maybe there is still time to pull this in for 1.8?

maxhartshorn:
@sboisvert @justnorris I'm curious what are the conflicts that need to be resolved before we can merge this branch. We are eagerly anticipating this feature and would love to see it in v1.8 or v1.9. Is there anything I can do to help get this code in there?

sboisvert:
I'm going to differ to @philipjohn who's leading the release. I think this should be good to merge after we do some testing (We would also need to clean up the merge conflict). 

@maxhartshorn would you be willing to take the lead on testing? If it's useful after testing locally we can fast track a version up to your VIP site so it can be battle tested a bit more. Let us know if that works for you.

maxhartshorn:
@sboisvert That would work. We would be happy to QA this feature if it will speed up deployment. Also I'm noticing I can't actually view the merge conflicts...is that because I don't have write access to the repo?

sboisvert:
> Also I'm noticing I can't actually view the merge conflicts...is that because I don't have write access to the repo?

That's weird. Do you mean in the web editor version only https://github.com/Automattic/liveblog/pull/337/conflicts or also via the command line / local Git install?

justnorris:
I merged Liveblog master back in - the PR should now be in sync. At least for a few days.

maxhartshorn:
Thanks @justnorris! @sboisvert please let me know if you can fast track a version to our VIP instance for testing

maxhartshorn:
Also one comment @justnorris. I don't want to hold up the pull request as we are eager to test on VIP, but have you given any thought to adding `headline`, `image`, and `publisher` values to `liveBlogUpdates`? Google's structured data testing tool marks them as required attributes, though I don't know if it's enforced in practice

sboisvert:
@maxhartshorn Yup, just open a ticket about it or, even better, add it to your theme directly and reference this conversation.

justnorris:
I'll have another look at those fields Monday. The good news is that even if a field is missed or if you need to add more data to your LiveBlogPost - you can always use the `liveblog_schema_liveblogposting` filter to append more data.

maxhartshorn:
@sboisvert is there anything special I need to do to add this plugin to my theme? I find when I add it in my functions.php as an include I get a `fatal error: undefined class constant 'key'`. I don't see the error when I load it from outside the theme via the `wpcom_vip_load_plugin` function

`include __DIR__ . '/includes/plugins/liveblog-test/liveblog.php';`

sboisvert:
@maxhartshorn let's move this to a ZD ticket I think it'll be easier to manage :)

maxhartshorn:
@sboisvert just wanted to follow up on the status of this feature. I see that it has been merged, does that mean that it will be in the next release?

sboisvert:
@maxhartshorn I think that was an accident on @justnorris's part. Can you confirm @justnorris ?

maxhartshorn:
@sboisvert @justnorris @philipjohn we are currently testing this feature locally. So far the schema output looks fine and I am able to add additional schema fields via the filter `liveblog_schema_liveblogposting`. Pending some more QA we'd love to try this feature out on a live blog tomorrow to see if/how Google picks up the schema.

My question is, given that this feature branch is more or less up to date with master, is there a risk running it on production? For example if there are other untested features in master we would be introducing by testing this branch?

justnorris:
GitHub notifications have gotten to me too :)

@maxhartshorn Yes - it was merged in by accident and then reverted.

I don't think there is any risk of running it in production apart from it being untested. I think I'll coordinate with @philipjohn to get this merged in for real soon.

maxhartshorn:
@philipjohn @justnorris I've tested this branch locally and it works well. I am able to validate the schema properly using the Google schema validation tool. And I am able to add my own schema tags with `liveblog_schema_liveblogposting` filter

There are a few places in `class-wpcom-liveblog-schema.php` the code where the variable `key` is lower-cased which was causing errors for me as the variable is defined as `KEY`. Changing the variable name to uppercase fixed the error

Here is an example from L23 of `class-wpcom-liveblog-schema.php` 
`$this->query   = new WPCOM_Liveblog_Entry_Query( $post_id, WPCOM_Liveblog::key );`

I would recommend merging after these minor fixes as this is a great feature to have in an upcoming release.

PR #128, opened 2014-02-11
Add `Upload image` link for mobile.

Marius786:
This is to enable full mobile support for images.

PR #338, opened 2018-01-17
Don’t force mobile devices in capture mode

justnorris:
I think that there is really no point of forcing camera capture on people - workflows differ.
It’s reasonable to assume people take pictures, edit them on the fly and then upload. Another scenario would be a reporter working form an iPad would be now able to upload an attachment received by Email.

### DIFF
--- a/src/react/Editor/Editor.js
+++ b/src/react/Editor/Editor.js
@@ -342,7 +342,6 @@ class EditorWrapper extends Component {
           id={this.inputId}
           onChange={this.uploadImages.bind(this)}
           accept="image/jpeg,image/gif,image/png,image/jpg"
-          capture="camera"
         />
         <Toolbar
           imageInputId={this.inputId}


sboisvert:
Thanks @justnorris! 
Is there an issue linked to this? What benefits does having the capture camera give us? Does it make it easier to take pictures? Is it just as easy to take pictures without the capture argument?

justnorris:
There is no issue that would be exactly linked to this, but I was inspired by this comment in #128:

> Dear developer,i use liveblog for publishing text and photos from several events. It will be perfect, if a future version of liveblog contains a button "Select photos" and i can choose a photo from my mobile phone, like apple. Currently i have to use a laptop. And this is very cumbersome if you are sitting in a stadion.Thankx,Denis
--

Right now, when you click "Select Image" on mobile you camera mode is instantly activated. If the capture mode is disabled - you get a native dialog offering you to either open the camera or browse for photos.

PR #339, opened 2018-01-17
Fix #336 Gravatars

### LINKED ISSUE #336: "false" string appears when Gravatars are disabled
When Gravatars are disabled, instead of not showing anything - the plugin prints out "false" as a string in place of the avatar. 

Was reported as a side issue at #333

justnorris:
Fixes #336 by hiding Gravatar HTML when there is no avatar available.


justnorris:
Closing in favor of #340

Issue #333, opened 2018-01-15
nhour ago function seems wonky


TwoFlags:
Love this plugin. Seems to work fine, except for two minor issues:

The "nh ago" function shows an incorrect timeframe, even though the time the post was published directly below that is correct.

And there is a "false" notation beside the username slug. 

Both issues show here: https://www.bossiernow.com/bossier-now-live-coverage-winter-storm-2018/

Thanks!

justnorris:
Hey there,
Are there any steps I can take to reproduce an incorrect timeframe? Is it showing an incorrect timeframe only for hours, or minutes too ( for example, "8 minutes ago" )?

The "false" notation appears there because you have disabled Gravatars. I created a separate ticket ( #336 ) to track that, but a quick fix would be to just enable Gravatars.

TwoFlags:
Hi,

Thanks for getting back to me. It is not showing minutes, only hours —
beginning with "6 hours ago" regardless of when the post was published.


*Hal M. Bundrick*
Publisher/Editor
BossierNow.com
(318) 286-1997

justnorris:
I managed to reproduce the problem. I'll take a look and see if there is anything I can do to help.

It looks like the function calculating how long ago something was posted isn't respecting Wordpress timezones.

liam-defty:
Thanks for pointing this out @TwoFlags!

@justnorris I have created a pull request with a fix. You were exactly right, it wasn't respecting the timezones. We have also switched to using moment.js as it makes handling things like this much easier.

gerhardsletten:
To me it looks like pull #340 are not fixing this issue. I am on gmt + 1 reflected in the General settings, but in order to have correct times I need to change the gmt_offset in liveblog.php to 120 instead of 60. I think the problem is with entry_time which gets its value from [get_comment_date](https://developer.wordpress.org/reference/functions/get_comment_date/) that returns the local time of the comment

PR #340, opened 2018-01-17
Fixes for #333 and #336

liam-defty:
**#336**

Fixes for avatar rendering incorrectly when Avatar Display is disabled. This was happening when rendering an entry and rendering the autocomplete for mentions in the editor.

**#333**
Switched to using moment.js to handle date formatting and handle the UTC offset between timezones correctly.


liam-defty:
We also now respect the chosen date and time format from the panel when rendering the date/time an entry was posted.

jasonagnew:
@justnorris I believe the two tests are now failing due to a core WordPress change. WordPress now returns 401, instead of 403. Would you like us to include a fix for the tests in this PR? or would you like to tackle that separately? Thanks.

Issue #223, opened 2015-12-10
Add support for other post types


philipjohn:
Allow pages, and other post types to be used as Liveblogs.

shantanu2704:
Hi @philipjohn!
I was wondering if there are any strategies in place to add this feature. Adding support for other built-in post types looks straightforward so I wanted to clarify a couple of doubts.

When adding support for other post types, should all the post types be supported by default or should there be a centralized settings page that lists all the post types available(including custom post types) and allows the user to choose which post types to support.

justnorris:
Hey @shantanu2704 
I was actually thinking of picking this one up today. 

I think that because there is no general Admin UI yet, there is no point of adding a menu item with just a single option inside ( "Support for Page post type: Yes/No" ). On top of that - when a post type is supported it'll display the Liveblog section where Liveblog has to be enabled before it will work.

As a solution, I'm thinking of just adding pages as a supported post type and applying a filter, so that anyone who wants to further extend the plugin can just add a filter to add or remove support for certain post types.

shantanu2704:
Hey @justnorris 

I have already made the necessary changes to add support for pages and attachments. There isn't really much to it. Adding a filter to disable certain post types, like you are suggesting, was something I was thinking of as well. But since I am new here, I thought I would see if someone has a better idea.

justnorris:
Oh, that's great. In that case submit a PR so someone can look it over. 

I'm not sure though about having a Liveblog for attachments. Maybe just pages and the filter? If someone needs a Liveblog for Attachments then they could use the filter.

shantanu2704:
Hey @justnorris!

I have submitted a PR( #342 ), but two tests are failing and I am not able to figure out how. If you could maybe look into it, that would be really helpful.
Thanks!

GaryJones:
Closed with #539.

PR #342, opened 2018-01-18
Add support for other post types #223

shantanu2704:
1. Added a static variable for supported post types that can be filtered with the 'modify_supported_post_types' filter in the init() method.

2. Added support for 'pages' in the init() method.


liveblog.php
```diff
@@ -245,9 +245,9 @@ public static function init() {
 		/**
 		 * This allows the users to filter their supported post types.
 		 */
-		$this->supported_post_types = apply_filters( 'modify_supported_post_types', $post_types );
+		self::$supported_post_types = apply_filters( 'modify_supported_post_types', $post_types );
```
justnorris: Filters need a prefix

liveblog.php
```diff
@@ -240,6 +240,9 @@ public static function init() {
 		 * we can possibly introduce this to other post types later.
 		 */
 		add_post_type_support( 'post', self::key );
+		add_post_type_support( 'page', self::key );
+		add_post_type_support( 'attachment', self::key );
+		add_post_type_support( 'revision', self::key );
```
justnorris: Are attachment and revision necessary?

liveblog.php
```diff
@@ -454,7 +454,7 @@ public static function is_liveblog_post( $post_id = null ) {
 	 * @return bool
 	 */
 	public static function is_viewing_liveblog_post() {
-		return (bool) ( is_single() && self::is_liveblog_post() );
+		return (bool) ( ( is_single() || is_page() || is_attachment() ) && self::is_liveblog_post() );
```
justnorris: Is a live blog necessary for attachments?

justnorris: Does this take into account the custom post types that can be added through the filter?

liveblog.php
```diff
@@ -461,7 +461,7 @@ public static function is_viewing_liveblog_post() {
 	 * One of: 'enable', 'archive', false.
 	 */
 	public static function get_liveblog_state( $post_id = null ) {
-		if ( ! is_single() && ! is_admin() && ! self::$is_rest_api_call) {
+		if ( !is_singular() && ! is_admin() && ! self::$is_rest_api_call) {
```
justnorris: This will return true for all post types, including unsupported post types.

.gitignore
```diff
@@ -7,6 +7,7 @@ yarn-error.log
 .svn
 composer.lock
 vendor/
+/nbproject
```
justnorris: I recommend that you add the /nbrpoject to your global gitignore, that way you won't have to add it to the project. Or you can move it to where the `.idea` folder is ignored

liveblog.php
```diff
@@ -236,11 +238,19 @@ private static function register_embed_handlers() {
...
-		 * we can possibly introduce this to other post types later.
+		 * Add liveblog support to the 'post' and 'page' post type.
+		 */		
+		$post_types = array( 'post', 'page' );
+		
+		/**
+		 * This allows the users to filter their supported post types.
 		 */
-		add_post_type_support( 'post', self::key );
+		self::$supported_post_types = apply_filters( 'modify_supported_post_types', $post_types );
```
justnorris: Filters have to be prefixed. [Some useful info here](https://vip.wordpress.com/documentation/code-review-what-we-look-for/#prefixing-functions-constants-classes-and-slugs).

liveblog.php
```diff
@@ -451,14 +461,14 @@ public static function is_liveblog_post( $post_id = null ) {
 	 * @return bool
 	 */
 	public static function is_viewing_liveblog_post() {
-		return (bool) ( is_single() && self::is_liveblog_post() );
+		return (bool) ( is_singular() && self::is_liveblog_post() );
```
justnorris: This will return true for all post types, including unsupported post types. This needs more precision. Also, possibly a filter - not sure.

liveblog.php
```diff
@@ -451,14 +461,14 @@ public static function is_liveblog_post( $post_id = null ) {
...
 	}
 
 	/**
 	 * One of: 'enable', 'archive', false.
 	 */
 	public static function get_liveblog_state( $post_id = null ) {
-		if ( ! is_single() && ! is_admin() && ! self::$is_rest_api_call) {
+		if ( !is_singular() && ! is_admin() && ! self::$is_rest_api_call) {
```
justnorris: Same as above - needs more precision.

liveblog.php
```diff
@@ -57,6 +57,8 @@
 	public static $is_rest_api_call       	= false;
 	public static $auto_archive_days     	= null;
 	public static $auto_archive_expiry_key  = 'liveblog_autoarchive_expiry_date';
+	
+	public static $supported_post_types = array();
```
justnorris: Why is a static variable better here than a regular class variable?

shantanu2704: The init method is static. As far as I know PHP allow non-static members to be accessed in a static method.

liveblog.php
shantanu2704: is_singular will return true for all default single post types. 
Correct me if I am wrong, but for it to return true on a custom post type, the name of the custom post type needs to be sent as a parameter. 

Also, how do you think this conditional should be handled?

shantanu2704: https://codex.wordpress.org/Function_Reference/is_singular#Custom_Post_Types

liveblog.php
shantanu2704: I think we could use get_post_type() here.
https://developer.wordpress.org/reference/functions/get_post_type/

Something like:
...
if ( !is_singular( get_post_type( $post_id ) ) ... )

liveblog.php
justnorris: Ahh, you're right - the `is_liveblog_post()` method is static too and it's going to rely on this.

liveblog.php
justnorris: `is_singular` by default will check if any post type is singular, not just post or pages. So you have to check the `is_signular()` against all of the post types that support LIveblog.

liveblog.php
justnorris: `is_singular` will accept post types as the parameter. You can pass the static `$supported_post_types` variable to the function instead.

liveblog.php
```diff
@@ -245,7 +245,7 @@ public static function init() {
 		/**
 		 * This allows the users to filter their supported post types.
 		 */
-		self::$supported_post_types = apply_filters( 'modify_supported_post_types', $post_types );
+		self::$supported_post_types = apply_filters( 'liveblog_modify_supported_post_types', $post_types );
```
justnorris: Can you remove the `modify` from the filter? I think that would be better because it would almost look like the variable name we're already using: `liveblog_supported_post_types`

justnorris:
Changes looking good! I think we need just a few more tweaks and we're good to go.

shantanu2704:
Hey @justnorris, I was wondering if you had a chance to look at the tests that are failing. I could use a little help there

justnorris:
Have a look at the test logs - could it be that your tests issue is related to #344 ?

shantanu2704:
It could be. 
The two tests that are failing, expect the HTTP error code 403 but are getting the HTTP error code 401.

justnorris:
In that case, I think we can wait until the tests are pulled in the master branch - then you can update your pull request to make sure the tests pass.

shantanu2704:
That sounds good. Thanks, Norris!

shantanu2704:
Hey @justnorris!
#344  has been merged. Is there a way to check if the tests that were previously failing are still failing or do they pass?

justnorris:
You can pull the master branch in your pull request branch - that should update everything and make the tests run again.

justnorris:
I think this looks good! /cc @philipjohn

GaryJones:
This PR has a merge conflict.

liveblog.php
```diff
@@ -236,11 +238,19 @@ private static function register_embed_handlers() {
 	 */
 	public static function init() {
 		/**
-		 * Add liveblog support to the 'post' post type. This is done here so
-		 * we can possibly introduce this to other post types later.
+		 * Add liveblog support to the 'post' and 'page' post type.
+		 */		
+		$post_types = array( 'post', 'page' );
```
philipjohn: If I understand correctly, this will also add support for pages out of the box? I'd rather we just stick to posts, and allowing pages to be added using the filter.

shantanu2704:
Hey @philipjohn! Thanks for the feedback!
Sadly, the changes made for this PR are long gone.
So I have submitted a new PR - https://github.com/Automattic/liveblog/pull/539 - which, along with the changes from this PR, incorporates your feedback, re the post type `page`.

Closing out this PR in favour of #539 :)

PR #343, opened 2018-01-18
Editor Enhancements

liam-defty:
**Improvements to the Editor** 

- Easier creation of editable blocks when developing
- Image Block with access to the media library
- HTML Block

## Blocks

When developing, it is now easier to create a 'block' within the editor using a [Higher Order Component](https://reactjs.org/docs/higher-order-components.html). This will make it focusable, draggable and editable meaning you can create new block components that share all of this default functionality without duplicated code.

## Image Block

An image block is now inserted into the editor instead of just inserting an image. This gives a user access to choose an image from the media library, change an image, or upload one.

The image size inserted into the content will default to 'full'. But we have added a filter to override this:

```
add_filter( 'liveblog_default_image_size', function () {
	return 'medium';
} );
```

## HTML Block

Rather than have a full HTML Editor which can cause issues. A HTML Block gives the user the ability to add custom HTML without compromising the visual editor. This could be useful for adding more complex markup such a tables, meaning editors are not limited to the features available of the editor.


philipjohn:
In testing this, when I try to upload an image, it seems to do nothing. Steps are;

1. Click on Insert Image
1. In the box that appears, click on "Upload an image?"
1. Choose an image from the local machine
1. "No images found" message has changed to "No Image Selected" bu there are no images to select 

This happened with both a gif and a jpg

Here's a screenshot:
fireshot capture 3 - liveblog wordpress dev - https___localhost_2018_01_24_liveblog_

liam-defty:
@philipjohn When the block shows 'Add Media' it should pull in images from the media library. If you have a fresh install perhaps you have no images in your database? However, if the image is failing to upload also then it sounds like there is an issue. I am unable to recreate the issue. Are there any errors in the console or have any network requests failed?

philipjohn:
Bingo. All works swimmingly with the media library working :smile:

philipjohn:
@liam-defty Any thoughts on the CI errors? Looks like an issue with NPM?

jasonagnew:
@philipjohn I've fixed those. I think when master was merged back in caused some of the styles to fail linting, was mainly single vs double quote, or a double definition of css property.

Issue #326, opened 2017-12-22
Gutenberg integration


danielbachhuber:
As a Liveblog user, it would be helpful if I could enable liveblog features while in the Gutenberg editor.

philipjohn:
Thanks for raising this! It's been on my mind, and we've already begun playing around with Liveblog and Gutenberg to see what's involved in adding support. Based on that, we're aiming to add Gutenberg support to v1.9.

We're going to look at;

- [ ] Adding a block to insert a Liveblog into a post
- [ ] Remove the meta box from the classic meta box
- [ ] Replace the classic meta box with an "Insert Liveblog" button above the editor, or an option within "Add Media"
- [ ] Possibly retire the templates feature, after a review of it's usage amongst users

jrmd:
Hey,

I have recently started working on this and I have it available to view here: https://github.com/bigbitecreative/liveblog/tree/feature/gutenberg-support

Although it's not quite finished, it does work fine from my testing, but I believe it requires some tidying up.

If anyone would like to look through/fork it and make changes I'd be grateful.

Thanks,
Jerome

jrmd:
Hey @danielbachhuber 

No problem, here is a small video of the implementation: http://bigbite.im/P8ZMIb

I thought all of the buttons made the block look quite large so i made it so they're only visible when you focus on the block.

jasonagnew:
Just to outline how this works. Both the Gutenberg block and Classic editor Insert Liveblog button place a shortcode into the content, this is then used to output the markup that React hooks into. To support older posts which don't have the shortcode as they didn't need too due to the post content filter, a piece of post meta is added on liveblogs posts created using the new button or block. If this meta is not there and the post is a liveblog then we run the old post content filter.

danielbachhuber:
Interesting! Would it eventually make sense to display the actual liveblog within Gutenberg?

PR #350, opened 2018-01-31
Gutenberg Integration

liamdefty:
As requested in #326 this provides support for Gutenberg.

The Gutenberg block inserts a short code into the content that outputs the required markup for the plugin to work instead of using the content filter. For backwards compatibility we add a piece of metadata when a short code is added. If the meta is not there and the post is a Liveblog then we run the old content filter. We hook into the save function to ensure the Liveblog metadata is correctly updated when the short code is changed. For support with the classic editor we have added an 'Insert Liveblog' button that will insert the short code.  

This caused a couple of issues:

1. When visiting the panel of a Liveblog post created before the Gutenberg integration the post would not show the Liveblog block in Gutenberg or the short code in the classic editor because it never existed. To fix this when a user visits this we use the `content_edit_pre` filter and `the_post` action (it seems the `content_edit_pre` filter no longer runs on Gutenberg) to output the short code/block in the content.

2. Switching to using a short code meant we lost the ability to be able to choose the key event formatting. So now we look to see if a post has the format meta, if it exists we use that key event format. Otherwise use a default that can be changed with a WordPress filter:

```
add_filter('liveblog_key_format', function() {
	return 'full';
});
```


philipjohn:
#594 supercedes this, with a more detailed list of requirements for a Gutenberg experience

PR #354, opened 2018-02-12
Use get_rest_url()

sboisvert:
By using get_rest_url() we can leverage the existing filters there and make sure we're consistent in the usage of REST API url.

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -1581,7 +1581,7 @@ final class WPCOM_Liveblog {
 	 * Checks if use_rest_api is on and the WordPress version supports it
 	 */
 	public static function use_rest_api() {
-		return apply_filters( 'liveblog_use_rest_api', self::use_rest_api && self::can_use_rest_api() );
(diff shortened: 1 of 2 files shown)


liveblog.php
```diff
@@ -1581,7 +1581,7 @@ public static function can_use_rest_api() {
 	 * Checks if use_rest_api is on and the WordPress version supports it
 	 */
 	public static function use_rest_api() {
-		return ( self::use_rest_api && self::can_use_rest_api() );
+		return apply_filters( 'liveblog_use_rest_api', self::use_rest_api && self::can_use_rest_api() );
```
philipjohn: We really want people to be using the REST API for the new React frontend. The plan is to completely deprecate the AJAX-based fallback, so giving folks the option to disable the REST API is going to force us into a position of breaking backwards compatibility. Let's leave this out for now, pending any strong needs-based arguments for it's inclusion.

PR #366, opened 2018-02-26
add ability to filter authors

paulschreiber:
This lets you filter author and contributor lists so you can populate them from another data source (such as CoAuthors Plus).


paulschreiber:
Additional work is needed to make the gravatars work for filtered authors.

paulschreiber:
I created a `get_avatar()` helper method, and replaced the calls to `get_avatar()` with calls to my helper. I added an `liveblog_author_avatar` filter to this helper.

Issue #371, opened 2018-03-13
Issue with update live blog Entry


achaitanyajami:
Hi Automattic,

I have observed a strange behavior in the Liveblog plugin when an update I made to live blog entry.
Here is the screenshot.
<img width="848" alt="screen shot 2018-03-13 at 7 09 29 pm" src="https://user-images.githubusercontent.com/27680483/37345097-219132fe-26f2-11e8-8ec7-3cd4c14fa6fd.png">

I can see this issue with the latest version as well.

Here is the how I can reproduce the issue:
* Create An Live blog entry.
* Edit the same LiveBlog entry and add extra content to it and update the content.
* Now delete the same entry which you have updated.
* Now you will see content is deleted without any alert information where we usually get when deleting the content.
* Now reload the page. You will see the above screen.

So, I have gone through the code and found that we are not checking if an entry exists in the DB.
In `liveblog/classes/class-wpcom-liveblog-entry.php`  function `get_fields_for_render` Line 97
`$entry_id     = $this->replaces ? $this->replaces : $this->comment->comment_ID;` not checking if this`$entry_id`  ID exists or not.

So I simply added this 2 lines of code in `liveblog/classes/class-wpcom-liveblog-entry.php`  function `render` after Line 130
`if ( ! wp_get_comment_status( $entry['entry_id'] ) ) {
			return $output;
		}`

So that I can only display/loop the file only if an entry ID exists.

Hope this will be Helpful, Sorry that I don't have permission to create a PR for this fix.

rebeccahum:
@achaitanyajami Thanks for reporting this issue!  

> Hope this will be Helpful, Sorry that I don't have permission to create a PR for this fix.

You should definitely be able to make a PR.  Do you want to try to give that a shot?

achaitanyajami:
Yes, @rebeccahum

rebeccahum:
Awesome! Looking forward to seeing your PR @achaitanyajami.

achaitanyajami:
Small change in the code,
 since `wp_get_comment_status ` only checks for 
`if ( $approved == null )
		return false;
	elseif ( $approved == '1' )
		return 'approved';
	elseif ( $approved == '0' )
		return 'unapproved';
	elseif ( $approved == 'spam' )
		return 'spam';
	elseif ( $approved == 'trash' )
		return 'trash';
	else
		return false;`

I have changed the code to `$comment = get_comment( $entry['entry_id'] );
[...]

PR #373, opened 2018-03-19
Display ENTRY If ID exists in the Database Else Return

arunchaitanyajami:
Hi Automattic,
Here is the PR i have created for the issue : 
https://github.com/Automattic/liveblog/issues/371


okupdatenumber:
https://github.com/Automattic/liveblog/issues/656#issue-589942960

GaryJones:
The issue in #371 was unable to be reproduced, and we've moved on several versions since then, so I'm closing this one out. 

If you are still able to reproduce it with the latest 1.9.7, we can re-open and test this again.

Issue #372, opened 2018-03-16
As of v1.7 you can no longer specify a custom key events title


maxhartshorn:
The documentation states that you can customize the title of the key events box by adding a title attribute to the shortcode:

[liveblog_key_events title="My New Title"]

This feature is no longer available in v1.7 as the title "Key Events" is hard-coded in the React template.

Is it possible to add this customization back in to the next release? Otherwise I would suggest updating the documentation as it is confusing.

lidocaine:
It appears that somewhere between v1.6.1 and v1.7, the content of the Key Events widget moved to being set by React. In `src/react/containers/EventsContainer.js:18`, the title has been hardcoded as Key Events.

I will see if I can create a workaround to pass the custom title to React.

lidocaine:
I took a look and originally tried to figure out a way to pass the custom title data into React; however, in the end, I ended up removing the `<h2>` from the React template and placing it back into `templates/liveblog-key-events.php`.  I've pushed these changes into PR #381; let me know if this works for you, @maxhartshorn

philipjohn:
Fixed by #381

PR #381, opened 2018-03-21
Fix for issue 372 - Key events title was hardcoded in react container

lidocaine:
Related Issue: #372 

When the React front-end elemens moved to React, the Key Events container title was hardcoded inside `src/react/containers/EventsContainer.js` which disabled the custom title attribute in the shortcode from functioning.

I'm not very familiar with React and after investigating possible solutions to pass the custom title to React, I settled on removing the hardcoded title from EventsContainer.js and putting it back in `templates/liveblog-key-events.php` along with the .widget-title class. This restored functionality and maintains the style.  Though it might make sense pass a title-class argument to the shortcode so to make its use outside of a widget area more flexible.


sboisvert:
Thanks for the PR @lidocaine!
I had a few questions:

Do we lose any functionality changes by doing it this way?

>Though it might make sense pass a title-class argument to the shortcode so to make its use outside of a widget area more flexible.

In what circumstances are we potentially restricted by this change?

cc @philipjohn who might have thoughts on the matter :)

lidocaine:
> Do we lose any functionality changes by doing it this way?

- Styles originally targeting the old `<h2>` might need be updated as it has been moved outside of the containing `<div>` which hold the actual list of events.
- We lose the dynamic-nature of React injecting the element into the DOM. Honestly this solution isn't exactly _ideal_ as React was injecting all of the list data into the `<div id="liveblog-key-events">` element, which was a clear separation of relevant elements. Now, with the `<h2>` outside of the React container, it can be a bit confusing at the very least.
- Originally, when all the Key Events data was in the `liveblog-key-events.php` template, there was no class on the `<h2>` and with my change, `.widget-title` is (still) hardcoded into the template.

*Note:* This morning I (finally) figured out how I could pass the custom title to React via a `data` attribute in the parent `<div>` . This is how I would handle something like this when using jQuery, but I'm not sure if this is within React's best practices.

> In what circumstances are we potentially restricted by [passing a title-class argument to the shortcode]?

I'm not sure; I can't think of anything beyond that it would be providing users with something superfluous as no one has actually asked for this functionality.

sboisvert:
> Note: This morning I (finally) figured out how I could pass the custom title to React via a data attribute in the parent <div> . This is how I would handle something like this when using jQuery, but I'm not sure if this is within React's best practices.

@jacklenox or @philipjohn might have input on this :)

sboisvert:
also cc @scottsweb

lidocaine:
I've updated my fix to pass the custom title as a `data` attribute in `div#liveblog-key-events` and handle the displaying of the `h2` in the React container, `EventsContainer`.

lidocaine:
Didn't realize use of `findDOMNode` was discouraged which (along with some minor syntax issues) caused the build to fail. I've cleaned up the code and avoided use of `findDOMNode` in most recent commit.

tomjn:
@lidocaine I would avoid needing to use `setState`, and instead pass the custom title in as a prop from 1 level up. That would eliminate needing to pass the container dom element to the component entirely.

Additionally, `EventsContainer.propTypes = {` needs modifying, changing to a prop and passing the title rather than a DOM element should make this bit easier

lidocaine:
> ...instead pass the custom title in as a prop from 1 level up. That would eliminate needing to pass the container dom element to the component entirely. Additionally, EventsContainer.propTypes = { needs modifying...

I think I did this right @tomjn -- Per my most recent commit, I'm passing the title as a prop from `AppContainer` and modified `EventsContainer.propTypes = {` to include `title` as `PropType.string`

cain:
Would love to see this merged!

Issue #370, opened 2018-02-26
Ability to edit HTML for posts


paulschreiber:
Users should be able to edit HTML for live blog posts, as they can with regular WordPress updates.

philipjohn:
Fixed by #383

Issue #368, opened 2018-02-26
Allow posts without authors


paulschreiber:
Allow posts without an author. Currently, attempting this fails silently.

In the JavaScript console, you see:
```
Uncaught TypeError: Cannot read property 'id' of undefined
```
`1.8 beta 1`

philipjohn:
Fixed by #383

Issue #367, opened 2018-02-26
Simplify authors (multiple authors, no contributors)


paulschreiber:
We want updates to have:
* multiple authors
* only one UI field (no distinction between authors and contributors)

philipjohn:
Fixed by #383

PR #383, opened 2018-03-23
1.8 changes

liam-defty:
## Raw Edit Feature

As requested in #370 this pull request contains the ability to be able to edit the html of an entry in the editor. As the editor is built with Draft.JS it will try to keep the two in sync. For more complex markup that is inserted in raw edit mode that the editor would not be able to render without custom blocks being created, such as tables, it will insert this into the new code block in the editor.

## Authors

The author UI is now simplified like suggested in #367 and posts can now have no authors like suggested in #368.

The author list is now able to be customised by building upon the pull request  #366. This allows support for integrating with things such as the co authors plugin. Here is an example of how this could work: 

```php
add_filter( 'liveblog_author_list', function( $users, $search ) {
    global $coauthors_plus;

    $args = array(
        'orderby' => 'name',
        'hide_empty' => false,
        'number' => 20,
        'search' => $search,
    );

    $author_terms = get_terms( $coauthors_plus->coauthor_taxonomy, $args );

    return array_map( function( $author ) {
        return (object) [
            'ID' => $author->term_id,
            'user_nicename' => $author->name,
            'display_name' => $author->name,
        ];
    }, $author_terms );
}, 10, 2 );

add_filter( 'liveblog_userdata', function( $userdata, $id ) {
    if ( $userdata ) {
        return $userdata;
    }

    global $coauthors_plus;

    $author = get_term( $id, $coauthors_plus->coauthor_taxonomy );

    return (object) [
        'ID' => $id,
        'user_nicename' => $author->name,
        'display_name' => $author->name,
    ];
}, 10, 2 );
```


classes/class-wpcom-liveblog-entry.php
```diff
@@ -358,45 +354,56 @@ public static function handle_restricted_shortcodes( $args ) {
...
 	 * @return mixed
 	 */
 	private static function handle_author_select( $args, $entry_id ) {
 		if ( isset( $args['author_id'] ) && $args['author_id'] ) {
-			$user_object = get_userdata( $args['author_id'] );
+			$user_object = self::get_userdata_with_filter( $args['author_id'] );
 			if ( $user_object ) {
 				$args['user'] = $user_object;
 
-				if ( $entry_id ) {
```
philipjohn: What's the reason for removing the conditional here? Looks like this is probably to protect against errors if `$entry_id` isn't set.

jrmd: @philipjohn I believe the if statement was removed due to it being redundant as all calls to the function make a check to see if the `$editor_id` parameter is set and not empty prior to calling this function; in addition to this the aforementioned if statement is wrapped within another if statement meaning if the earlier if fails then it resorts to using the `else` block.

As a solution we could superfluously add this if statement back but at the top of this function block which will just return `$args['user'];` to prevent the rest of the code running.

Issue #362, opened 2018-02-25
Filters not firing


philsbury:
Hi,

I've tried using two of the filters in the plugin, one to add a custom key event just based on the docs and on to show more in the feed, but neither seem to even fire. I've used my own theme and twentyseventeen, also class and non class based application of the filters and nothing seems to work. Am I missing something? Here's what I've tried:

```php
add_filter('liveblog_number_of_default_entries', 'default_entries', 10);

function default_entries(){
  error_log('updating default entries'); // no output 
  return 10;
}

add_filter( 'liveblog_active_commands', 'add_highlight_command', 10 );

function add_highlight_command( $commands ) {
  error_log('highlight'); // no output
  $commands[] = 'highlight';
  return $commands;
}
```

Thanks

philipjohn:
Fixed by #384 but shout if you're still having issues.

PR #384, opened 2018-03-25
Apply filter `liveblog_active_commands` later on so that hooking works

mikeyarce:
Currently, the filter `liveblog_active_commands` runs early becaue it's being loaded when the plugin class is instantiated.   If you happen to hook into it with a plugin that loads before `liveblog` does, then it does work, but if your plugin loads after `liveblog` then the filter won't apply.

This PR moves the filter to `after_setup_theme` so that it waits for all possible plugins and other functions to load first and then it runs the filter - that makes sure that your hook will be used!

This also updates the Readme with updated instructions.
Fixes second part of #362


sboisvert:
@mikeyarce Wanna solve the conflict here?
@philipjohn Do you want to give a review and merge if you're good with it after the merge conflict is done?

mikeyarce:
@sboisvert added a better commit message!

sboisvert:
Ugh @philipjohn I didn't mean to merge I just wanted to fix the conflict but the UI wasn't super clear :(

Can you confirm if this is good or if I should revert it?

combs:
hi when can this be merged in? We are stuck waiting for it in our production environment; we need to disable hashtag parsing since it is misfiring wildly on numeric entities for us. (#418)

sboisvert:
@philipjohn I'm going to go ahead and merge this but feel free to revert if you disagree.

Thanks!

PR #394, opened 2018-04-10
work around get_comment_date() GMT offset bug

### LINKED ISSUE #378: entry_time sent in local time over JSON (instead of GMT)
In the JSON data (i.e. `/wp-json/liveblog/v1/122233/get-entries/1/11-1521657625/`), `entry_time` is sent across the wire in local time.

This comes from `classes/class-wpcom-liveblog-entry.php`'s `for_json()` function:
```php
'entry_time'   => get_comment_date( 'U', $entry_id ),
```

In WordPress core, in `get_comment_date()` is found in `wp-includes/comment-template.php`:

```php
function get_comment_date( $d = '', $comment_ID = 0 ) {
  $comment = get_comment( $comment_ID );
  if ( '' == $d ) {
    $date = mysql2date( get_option( 'date_format' ), $comment->comment_date );
  } else {
    $date = mysql2date( $d, $comment->comment_date );
  }
```

unlike `get_comment_time()`, which can return both GMT and non-GMT dates, `get_comment_date()` can only return non-GMT dates.

In `EntryMeta.js`, the date is formatted with JavaScript:
```
formattedTime(entry.entry_time, config.utc_offset, config.time_format)
```

The `utc_offset` is set in `liveblog/liveblog.php`:
```
'utc_offset' => get_option( 'gmt_offset' ) * 60, // in minutes
```

`formattedTime()` assumes its input is in GMT/UTC. However, it is actually in local time (GMT -04:00 for us). This means the output of formattedTime is neither GMT _nor_ local time, but an incorrect value. In our case, we saw 2am instead of 2pm.

The workaround/fix for this is to double the offset.
```
'utc_offset'  => get_option( 'gmt_offset' ) * 120, // in minutes
```

A second workaround/fix would be to replace the `get_comment_date()` call with your own function that fetches the comment date in GMT.

paulschreiber:
Fixes #378.


paulschreiber:
Use #432 instead.

PR #420, opened 2018-04-30
update get_authors() to only return contributors

paulschreiber:


paulschreiber:
@philipjohn

philipjohn:
Thanks Paul - can you add details about what issue this addresses please?

PR #428, opened 2018-05-01
Adds check around the jetpack is_mobile flag

scottblackburn:
Adds conditional around jetpack `is_mobile` flag as having jetpack installed was assumed.

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -1089,7 +1089,7 @@ if ( ! class_exists( 'WPCOM_Liveblog' ) ) :
 			$settings = array(
 				'defaults' => $defaults,
 				'browser'  => array(
-					'mobile'    => jetpack_is_mobile(),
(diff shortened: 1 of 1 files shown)


justnorris:
Hey Scott,
Maybe instead of falling back to `false` the check can fall back to `wp_is_mobile()` instead?

liveblog.php
```diff
@@ -1089,7 +1089,7 @@ private static function add_default_plupload_settings() {
 			$settings = array(
 				'defaults' => $defaults,
 				'browser'  => array(
-					'mobile'    => jetpack_is_mobile(),
+					'mobile'    => ( function_exists( 'jetpack_is_mobile' ) ? jetpack_is_mobile() : false ),
```
paulschreiber: Add a `// phpcs:ignore` comment to the end of this line.

PR #432, opened 2018-05-02
Fix/gmt offset comments

scottblackburn:
This is a fix for #378 :

This now passes UTC time instead of WordPress time of entry to the frontend. From there we compare UTC to UTC for doing time ago. And entry date is worked out by taking UTC and adding/subtracting UTC offset.

I believe #394 was intended to fix this but this should address both the time ago and the entry time.


classes/class-wpcom-liveblog-entry.php
```diff
@@ -106,6 +106,22 @@ public function get_timestamp() {
...
+	/**
+	 * Retrieve the comment date of the current comment using gmt.
+	 * @param string          $d          Optional. The format of the date. Default user's setting.
+	 * @param int|WP_Comment  $comment_ID WP_Comment or ID of the comment for which to get the date.
+	 *                                    Default current comment.
+	 * @return string The comment's date.
+	 */
+	public function get_comment_date_gmt( $d = '', $comment_ID = 0 ) {
+	        $comment = get_comment( $comment_ID );
+	        if ( '' == $d )
```
paulschreiber: PHPCS notes:
* wrap the if/else statements in curly braces
* use $commend_id instead of $comment_ID
* Add spaces inside parentheses

paulschreiber:
This is a better approach than what I did in #394.

classes/class-wpcom-liveblog-entry.php
```diff
@@ -120,7 +136,7 @@ public function for_json() {
 			'css_classes' => $css_classes,
 			'timestamp'   => $this->get_timestamp(),
 			'authors'     => self::get_authors( $entry_id ),
-			'entry_time'  => get_comment_date( 'U', $entry_id ),
+			'entry_time'  => $this->get_comment_date_gmt( 'U', $entry_id ),
```
paulschreiber: Should `get_fields_for_render()` be updated to use this helper, too?

scottblackburn: Just looking into this, I don't think this function `get_fields_for_render()` is used anymore as the rendering is now done with react but it might be worth leaving this function in if rendering in php is required again in the future, the `for_json` function is used for the react render. I've also added phpcs fixes in.

paulschreiber: I was wondering why markup is sent over the wire and not used. It would be great to remove that now-dead code.

philipjohn:
Tested this just now, confirmed the bug and that this PR fixes the issue.

PR #433, opened 2018-05-02
Adds casting of last entry id and entry id when getting pages entries

### LINKED ISSUE #392: Pagination Does Not Update Properly
## Steps to Reproduce
1) Add 6 liveblog entries consecutively without refreshing or reloading the page

## Expected Results
Pagination should update to "1 of 2" and clicking "Next" button should bring you to the next page

## Actual Results
Pagination remains at "1 of 1" and clicking "Next" does not do anything
image

scottblackburn:
Some PHPCS changes to the `get_entries_paged` function seem to have caused issues with the pagination, this PR just casts the id's to `int` so that the strict param in `array_search` works as expected. This may also resolve #392 but needs checking.

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -807,7 +807,7 @@ if ( ! class_exists( 'WPCOM_Liveblog' ) ) :
 			if ( $last_known_entry ) {
 				$last_known_entry = explode( '-', $last_known_entry );
 				if ( isset( $last_known_entry[0], $last_known_entry[1] ) ) {
-					$last_entry_id = $last_known_entry[0];
+					$last_entry_id = (int) $last_known_entry[0];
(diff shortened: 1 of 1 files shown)


liveblog.php
```diff
@@ -807,7 +807,7 @@ public static function get_entries_paged( $page, $last_known_entry = false, $id
 			if ( $last_known_entry ) {
 				$last_known_entry = explode( '-', $last_known_entry );
 				if ( isset( $last_known_entry[0], $last_known_entry[1] ) ) {
-					$last_entry_id = $last_known_entry[0];
+					$last_entry_id = (int) $last_known_entry[0];
 					$index         = array_search( $last_entry_id, array_keys( $entries ), true );
```
paulschreiber: Should the cast go in the `array_search()` call to match the below example?

PR #439, opened 2018-05-09
Cast liveblog rewrite version before checks

mjangda:
The option is returned is a `string` and the class holds the version as an `int`. The strict check causes the version check to fail and we end up flushing rewrites every time `init` fires.

Before doing the check, cast the rewrite version option as an `int` to avoid the type mismatch.


mjangda:
Bug was introduced here: https://github.com/Automattic/liveblog/commit/5be98e929fb10550f06fc3aa1bd98c3318ddf43b#diff-d203c752daf3d7d646104737fd3a4c02R262

PR #442, opened 2018-05-09
Feature/load-more - Toggle lazy loading feature

cain:
screen shot 2018-05-09 at 4 51 03 pm


src/react/containers/AppContainer.js
```diff
@@ -61,8 +62,7 @@ const mapStateToProps = state => ({
   page: state.pagination.page,
   loading: state.api.loading,
   entries: Object.keys(state.api.entries)
-    .map(key => state.api.entries[key])
-    .slice(0, state.config.entries_per_page),
```
cain: I had to remove this slice method, seems to still work perfectly. Would love some feedback on why this was used.

cain:
Rebased off 1.8.2 master.

cain:
@philipjohn thoughts? 😸

cain:
Still TODO:
- More testing (clicking on key events)
- Check on review above

adamsilverstein:
> Change jump-to-key-event to contain all previous pages for load more functionality

@cain what is the " jump-to-key-event" this part refer to as missing?

cain:
@adamsilverstein 
So, when "jump-to-key-event" action is called, it will only load x number (entries per page) of entries before the desired "event". Ideally we want *all* entries *before* the event we are jumping to.

Eg. if the event is on page 3, load the first 2 pages PLUS the 3rd page, then scroll to event. Let me know if this doesnt make sense.

We pushed a feature to our local repo, but i havent pushed the code to github yet. Kind of tricky since  we created a new endpoint to help deal with this. Ill push the code soon.

cain:
@adamsilverstein 
Do you want to push your changes you made in your PR to this branch?

adamsilverstein:
> Ideally we want all entries before the event we are jumping to.

So would this mean 'load more' would load all remaining entries vs. another page of entries?

> Do you want to push your changes you made in your PR to this branch?

Sure, I'm not quite finished, I'll pull the changes back in here when I am.

I'm going to add logic to change the load more button into a spinner once you click it so you can tell something is happening (and to prevent clicking it twice). In addition, we are going to enable (via a filter) scroll detection that eliminates the Load More button entirely and instead lazy loads entries as your scroll approaches the bottom of the stream

cain:
>So would this mean 'load more' would load all remaining entries vs. another page of entries?

Sorry im confused by the question.
The "load more" button just gets the next page of data. I'm trying to say that, when using the "jump to event" action on load with the hash in the url it makes sense to load all PREVIOUS entries to the event. The current jump to event only load the number of entries in that page.

Eg: if my url is `website.com/wp-liveblog/123123/#12` and we have 5 entries per page, on load we want every entry before id 12 correct? So we had to create a new entries to get page1, page2, page3 all in one endpoint.

I dont think this is an issue for you guys since you load 50 entries per page.

>I'm going to add logic to change the load more button into a spinner once you click it so you can tell something is happening (and to prevent clicking it twice). In addition, we are going to enable (via a filter) scroll detection lazy loads entries as your scroll approaches the bottom of the stream

Sounds good! We arent interested in the infinite scroll at the moment, but would be a good feature to have!

adamsilverstein:
> The "load more" button just gets the next page of data. I'm trying to say that, when using the jumpToEvent action on load with the hash in the URL it makes sense to load all PREVIOUS entries to the event. The current jump to an event only loads the number of entries in that page.

Ah, ok - got it, i was confused.

fabianpimminger:
Any news on this?

adamsilverstein:
@fabianpimminger we implemented this feature for fivethirtyeaight's implementation, however I'm not sure these changes were ever merged upstream. Take a look at this PR for some details and in case you want to add the functionality to your site: https://github.com/paulschreiber/liveblog/pull/28

fabianpimminger:
Thx @adamsilverstein .. any chance this gets merged?

cain:
I'm no longer supporting this, just need to get it tested and merged by an admin.

Issue #449, opened 2018-05-11
Javascript Frontend Tests


cain:
Thoughts on creating some react tests? Possibly Jest? 😸

philipjohn:
Yes! Happy to hear suggestions, as a more comprehensive test framework is something we need to dedicate some time to.

PR #451, opened 2018-05-11
Feature/tests

scottblackburn:
As tests have been mentioned in #449, we thought it would be worth sharing these tests that were added for 1.7 and could be a good base for extending in the future.


cain:
Thanks @scottblackburn ill take a look at this! 😄

Issue #452, opened 2018-05-14
getImageThumbnail error breaks whole React component when images have missing data


jvpcode:
If a given site contains media library entries that are broken in some way, (i.e. they don't have media details or thumbnail sizes) a javascript error is thrown and the component breaks like so :

![medialibmaster](https://user-images.githubusercontent.com/19501762/39983343-b3d88a62-579a-11e8-82d4-448777bf73d6.gif)

cain:
+1

jvpcode:
Issue is specific to our wordpress install.

creativeagent:
Hi, thanks for reference this to my problem. is there a Solution?

cain:
Look at the conditional at https://github.com/Automattic/liveblog/pull/453/files#diff-02705067ba05489137df59bec522f5b7R17 

Either copy paste this line above into your codebase and rebuild the javascript files or pull the feature branch we created on our fork: https://github.com/newscorpaus/liveblog/tree/bugfix/insert_image_thumbnail_error_fix 

Let me know if you need help :)

PR #453, opened 2018-05-14
Additional conditionals to cater for images with missing data

jvpcode:
Additional conditionals to cater for images with missing data when getting thumbnails. The missing data otherwise breaks the whole liveblog: 

Addressing the following [issue](https://github.com/Automattic/liveblog/issues/452)

medialibmaster


philipjohn:
Is this still relevant @jvpcode ? I notice that #452 is closed with this comment: 

> Issue is specific to our wordpress install.

Issue #375, opened 2018-03-20
Add headline/title to each update


paulschreiber:
Allow live blog updates to have an optional headline/title.

cain:
@paulschreiber 
any specifications on the best way to do this? Inside the editor or a separate field? I love the way medium does it.

adamsilverstein:
@cain We implemented this as comment meta on the live updates. A new headline field is added above the editor and sent along with `crud` requests. You can see the changes for this feature (with only a few auxiliary changes) in this PR: https://github.com/paulschreiber/liveblog/pull/17/files and as Paul mentioned I'm working on a more complete PR with all of our recent updates.

Issue #464, opened 2018-05-24
Key events title uses anything


cain:
<img width="840" alt="screen shot 2018-05-23 at 2 57 21 pm" src="https://user-images.githubusercontent.com/15246256/40457792-c6b711aa-5f3b-11e8-8a9f-3a890ed59619.png">

The key event title just grabs the first piece of content in the post resulting in images or other content being used in the widget. #375 maybe?

philipjohn:
For this, I think we need to do two things:

1. Make the selection smarter, stripping out images etc before grabbing the first bit.
2. Add the ability to add a specific headline

cain:
@philipjohn For our fork, we require a key event to have a title. This makes it really simple for our editors ect... Have a look on the right hand col: 

https://www.heraldsun.com.au/sport/afl/teams/afl-grand-final-2018-live-collingwood-v-west-coast-eagles-at-the-mcg/live-coverage/e7de69f83d0cfa5c3572eb620ada3211

PR #473, opened 2018-05-29
Add entry headline

cain:
screen shot 2018-05-28 at 6 19 30 pm
screen shot 2018-05-29 at 11 13 21 am

This is just a pr I created from https://github.com/paulschreiber/liveblog/pull/17/files Just cleaned up a little and some styling changes that were specific to fivethirtyeight

Code stolen from @paulschreiber @adamsilverstein

PR #474, opened 2018-05-30
Add Entry Headline and clear when creating new entry

adamsilverstein:
Extends https://github.com/Automattic/liveblog/pull/473 with clearing of the headline editor when creating a new entry.

Fixes https://github.com/Automattic/liveblog/issues/375

Based on https://github.com/paulschreiber/liveblog/pull/17

* Add a headline field above the TinyMCE editor, (similar to LivePress).
* Headline field is optional.
* Store the headline in wp_comment meta.
* Send the headline over JSON as part of the API request for updates.
* Render the headline, if present, in EntryContainer.js
* php rendering for headline in  liveblog-single-entry.php
* Clear headline after posting from main/top editor (track last update)
* Sync headline from data when using inline editor/editing exiting entries


adamsilverstein:
Added the built files as well, @cain can you give this a test?

cain:
@adamsilverstein Done some basic testing, seems to be working fine. 

Whats everyones thoughts on using the headline as the key event widget content? Seems like it would be perfect and fix issue #464

philipjohn:
> Whats everyones thoughts on using the headline as the key event widget content? Seems like it would be perfect and fix issue #464

Yep, I think that'd be a good addition!

GaryJones:
This is going to need to be rebased to be testable.

PR #494, opened 2018-06-26
Add loading state to api actions

cain:
Added loading to some api actions


philipjohn:
Thanks for this - can you expand this PR with details about what issue this fixes with details reproduction steps please?

cain:
@philipjohn So in the state we hold a `loading` variable to handle various tasks while waiting for the api to return. This PR just expands `loading` state to the other redux api actions.

EG: Deleting/editing entries will now have a `loading` state to do conditionals off.

I'm not sure this is 100% neccessary if #527 is merged into master, since that PR uses a `isPublishing` state variable instead.

cain:
@davidsword is this PR neccessary or useless once your PR is into master :)

philipjohn:
You may be right about that, and I'm planning to test #527 and include it in v2.0 if all is well :)

cain:
Yeah i think that PR covers this. Would be good to get @davidsword's feedback

davidsword:
Hey! Sorry for the delay. This is a good idea and similar to what I was looking for before my PR. I like the naming 'loading' better.

I feel this lacks the specific of what is being done, and to what. In the situation I needed something like this for, <Entry> and <EditorContainer>, I needed to know if it was submitting a new entry, or when updating, which entry ID was being updated, which delete button to style differently, ect.

cain:
Yeah i agree with @davidsword 
Please feel free to open this back up if its necessary

Issue #492, opened 2018-06-21
App.js needs to be reduced (2.24 MB)


cain:
### Problem
App.js needs to be reduced for the functionality it provides.

### Solutions
- Multiple node modules need to be moved to dev dependencies.
- **Code splitting for authoring components, especially the EditorContainer (Big performance hit)**
- The `brace` module is just too big for what we are using it for (editor HTML blocks). Probably best to look for an alternate.
- Moment locale is quite large, and I don't think we are using it. Maybe https://www.npmjs.com/package/dateformat
- Immutable is being introduced in multiple locations.
- ~Extract vendors into a separate build file.~ (no need for this since its a single page)
- rxjs global import? https://github.com/Automattic/liveblog/blob/master/src/react/epics/index.js#L1

webpack bundle analyzer: 
**app.js = 2.19mb**
node_modules = 2.09mb
brace = 1.04mb
moment locale = 0.2mb
immutable.js = 50kb * 3
![untitled](https://user-images.githubusercontent.com/15246256/41699378-f1588394-7566-11e8-99de-1330958fb7fb.jpg)

If anyone else has any suggestions please let me know, ill be working on this 😄

philipjohn:
Fixed by #495

PR #495, opened 2018-06-29
Bundle size, improve performance.

cain:
### Whats been done:
- Code split EditorContainer into a separate bundle, we only want to load this if needed.
- Remove global import of Rxjs, only import what we need
- Moment locale is never used, removed on webpack build
- Set webpack_public_path
- update .eslint to use es2015

### Difference
Gzipped files reduced from 600kb to **90kb** for the client. Authoring bundle now is separate, so we don't care about the size.

PR created for issue #492


src/react/epics/index.js
```diff
@@ -1,4 +1,12 @@
-import 'rxjs';
```
cain: Remove global import, only import what we need

webpack.config.js
```diff
@@ -100,6 +101,7 @@ const webpackConfig = {
       __TEST__: JSON.stringify(process.env.NODE_ENV === 'test'),
       'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
     }),
+    new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
```
cain: We dont use timezones with locale, which is quite big

src/react/index.js
```diff
@@ -11,6 +11,9 @@ Polyfills();
 
 const store = configureStore();
 
+/* eslint-disable camelcase, no-undef */
+__webpack_public_path__ = `${window.liveblog_settings.plugin_dir}assets/`;
```
cain: Create a path for bundle files

src/react/components/Editor.js
```diff
@@ -0,0 +1,9 @@
+import Loadable from 'react-loadable';
+
+export default Loadable({
+  loader: () =>
+    import('../containers/EditorContainer'),
```
cain: Dynamically import editorcontainer

.babelrc
```diff
@@ -1,4 +1,4 @@
 {
-  "presets": ["env","react"],
-  "plugins": ["transform-object-rest-spread", "transform-object-assign"]
+  "presets": ["es2015","env","react"],
+  "plugins": ["transform-object-rest-spread", "transform-object-assign", "dynamic-import-webpack"]
```
cain: update babel

cain:
Need to rebase and keep PR up to date with master as its been sitting here for a while...

PR #496, opened 2018-07-10
Round out polling timestamp

mjangda:
Rather than a random scatter of timestamps, we can get a much higher cache hit-rate by forcing/normalizing multiple clients to make a request with the same timestamp.

### DIFF
--- a/src/react/services/api.js
+++ b/src/react/services/api.js
@@ -21,7 +21,13 @@ export function getEntries(page, config, newestEntry) {
 }
 
 export function polling(newestEntryTimestamp, config) {
-  const timestamp = getCurrentTimestamp() + config.timeDifference;
+  let timestamp = getCurrentTimestamp() + config.timeDifference;
+
+  // Round out the timestamp to get a higher cache hitrate.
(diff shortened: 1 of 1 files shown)


mjangda:
Lint issues have been fixed,

paulschreiber:
Per @rogertheriault, this can result in timestamps in the future. Change `Math.round()` to `Math.floor()`.

rogertheriault:
We've (@mjangda actually) noticed that ` + config.timeDifference;` can sometimes add a few minutes to the timestamp, if the page was cached. So that may be part of the problem when there's an active cache - it doesn't show up too well when just testing the individual requests. It might be necessary to drop that bit. Even without it, we've seen cases where there were still some future timestamps...

paulschreiber:
I was told to remove `+ config.timeDifference` and did in my copy. Let's do that here, too.

philipjohn:
I've pushed the removal of the `+ config.timeDifference` which looks good. I'm going to go ahead and merge seeing as we all seem to be on the same page on this.

PR #499, opened 2018-07-19
Make set_request_last_from_entries() static

philipjohn:
Deals with PHP error generated by line 255 where this function was called statically.


philipjohn:
Apparently I already did this in 484

PR #502, opened 2018-07-25
Wider retrieval of authors for liveblog attribution

jvpcode:
Code change to retrieve contributors, editors and authors for liveblog attribution


philipjohn:
Thanks for this - can you elaborate on the issue this PR addresses please, and why this change is needed?

Issue #506, opened 2018-07-29
LiveBlogPosting schema validation errors


philipjohn:
In 1.9rc there are some issue with the LiveBlogPosting schema - it doesn't validate with the following errors and warnings:

# Errors 

- [ ] publisher is not a known valid target type for the publisher property.
- [ ] @type Unspecified Type (The @type is required and cannot be an empty string.)
- [ ] The attribute publisher.itemtype has an invalid value
- [ ] image A value for the image field is required.

# Warnings

- [ ] The mainEntityOfPage field is recommended. Please provide a value if available.
- [ ] The coverageEndTime field is recommended. Please provide a value if available.
- [ ] The coverageStartTime field is recommended. Please provide a value if available.

GaryJones:
Reference: [`LiveBlogPosting` schema](https://schema.org/LiveBlogPosting).
Reference: [`liveBlogUpdate` schema](https://schema.org/liveBlogUpdate).

----

> publisher is not a known valid target type for the publisher property.
> @type Unspecified Type (The @type is required and cannot be an empty string.)
> The attribute publisher.itemtype has an invalid value

These are related to the same issue. The current entry-level `publisher` tag contains an empty `@type` (`liveblog` is the name of my local test site):

```
"publisher":{"@type":"","name":"liveblog"}
```

Changing it to `Organization`, (which is what the page-level tag has, which also uses the same `name`), works here.

```
"publisher":{"@type":"Organization","name":"liveblog"}
```

I'm also seeing a:
> logo	A value for the logo field is required.
error, as my local test site doesn't have an image set.

But, like below, _schema.org does not require fields; schema.org data consumers require fields_.

----

> image A value for the image field is required.

Relevant: https://stackoverflow.com/questions/35710385/image-missing-and-required-wordpress-amp-structure-doesnt-add-image-attribute/35711711#35711711

> "If the Google SDTT says that a property is "missing and required", it does not mean that it’s required by AMP or Schema.org. It [only means](https://stackoverflow.com/a/35629108/1591669) that Google won’t show one of their Google Search results features (like Rich Snippets) for your page."

So, if a post has no image available, then that's fine. Default images [should not](https://github.com/ampproject/amphtml/issues/5404#issuecomment-251818433) be used.

One thing Genesis Framework does, if no featured image is set, is to parse the content and use the first image in the entry / first image _attached_ to the entry as the featured image on archives. This may or may not be desirable behaviour here.

----

[...]

PR #508, opened 2018-08-07
Schema: Use @type for publisher type

GaryJones:
The array key set in AMP (https://github.com/Automattic/amp-wp/blob/9bd2a05d62315738952bd95ac413b1983bf91423/includes/amp-helper-functions.php#L745) is `@type`, not `type`.

See #506.


GaryJones:
While this fixes the immediate bug, the block of logic here can be refactored into a much better state, and the bug fix will be incorporated there instead.

Issue #505, opened 2018-07-29
LiveBlogPosting schema not included in non-AMP liveblogs


philipjohn:
The addition of the LiveBlogPosting schema in [1.9rc](https://github.com/Automattic/liveblog/releases/tag/1.9-rc1) adds the metadata only in the AMP version of liveblogs.

The schema isn't an AMP-specific thing so we should add it to the non-AMP version as well.

GaryJones:
[Add filter](https://github.com/Automattic/liveblog/blob/fd833bb0d5788b13e5fc421641491fb185393024/classes/class-wpcom-liveblog-amp.php#L45), and the [callback](https://github.com/Automattic/liveblog/blob/fd833bb0d5788b13e5fc421641491fb185393024/classes/class-wpcom-liveblog-amp.php#L182-L236).

GaryJones:
Self notes:

- The [`class-wpcom-liveblog-entry.php`](https://github.com/Automattic/liveblog/blob/fd833bb0d5788b13e5fc421641491fb185393024/classes/class-wpcom-liveblog-entry.php) no longer appears to be in use.
- The non-AMP template appears to be [built](https://github.com/Automattic/liveblog/blob/fd833bb0d5788b13e5fc421641491fb185393024/src/react/containers/EntryContainer.js#L70-L116) in React.
- Not sure if there's a way to have the creation of the metadata reusably centralised, and then used via React and AMP outputs.

PR #509, opened 2018-08-07
AMP schema refactor

GaryJones:
This is the first step of fixing some of the issues in #506, and refactoring the code to aid with #505.

See individual commits for more info.

### DIFF
--- a/classes/class-wpcom-liveblog-amp.php
+++ b/classes/class-wpcom-liveblog-amp.php
@@ -198,16 +198,17 @@ class WPCOM_Liveblog_AMP {
 
 		foreach ( $entries[ 'entries' ] as $key => $entry ) {
 			$blog_item = [
-				'@type'         => 'BlogPosting',
-				'headline'      => self::get_entry_title( $entry ),
-				'url'           => $entry->share_link,
-				'datePublished' => date( 'c', $entry->entry_time ),
-				'dateModified'  => date( 'c', $entry->timestamp ),
-				'author'        => [
+				'@type'            => 'BlogPosting',
+				'headline'         => self::get_entry_title( $entry ),
+				'url'              => $entry->share_link,
+				'mainEntityOfPage' => $entry->share_link,
+				'datePublished'    => date( 'c', $entry->entry_time ),
+				'dateModified'     => date( 'c', $entry->timestamp ),
+				'author'           => [
 					'@type' => 'Person',
 					'name'  => $entry->authors[ 0 ][ 'name' ],
 				],
-				'articleBody'   => [
+				'articleBody'      => [
 					'@type' => 'Text',
 				],
 			];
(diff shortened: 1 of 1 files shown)


classes/class-wpcom-liveblog-amp.php
```diff
@@ -188,50 +188,41 @@ public static function append_liveblog_to_metadata( $metadata, $post ) {
...
+				],
+			];
+
+			if ( isset( $metadata['publisher'] ) ) {
+				$blog_item['publisher'] = $metadata['publisher'];
 			}
 
-			$metadata['@type']          = 'LiveBlogPosting';
-			$metadata['liveBlogUpdate'] = $blog_updates;
+			$blog_updates[] = json_decode( json_encode( $blog_item ) );
```
philipjohn: Can you elaborate on the need for this change please? I'm not sure I understand what it adds.

philipjohn:
I decided to tackle the abstraction myself and opened https://github.com/GaryJones/liveblog/pull/1

classes/class-wpcom-liveblog-amp.php
GaryJones: The previous code was casting each array to an `(object)`, which makes adding and amending data in the array (such as a conditional publisher logo) awkward. This conversion should be a one-time recursive task. `json_decode( json_encode() )` is a short-hand way of doing just that. This could be pulled into a separate method (such as `convertToObjects()`, or a better domain-related name) once the abstraction was done.

GaryJones:
> what can we do here to make sure that this code produces expected results?

Unit and integration tests would be the obvious solution, though `append_liveblog_to_metadata()` is not the test-friendliest of methods.

Test written before, shown to produce known results, then refactor and shown to produce desired results (typically the same, depending on the refactor).

> is your intention to develop this further to abstract this out for use in both AMP and non-AMP liveblogs

Yes - this was fixing some of the validation errors (#506), and would then have a separate PR for doing the abstraction.



> I decided to tackle the abstraction myself and opened GaryJones#1

Thanks! - I left one or two comments, and I'd want to review it again after it was merged into my branch.

I'd consider the addition of this schema metadata to be a decorator - that is, the liveblog would work fine without it, but this just enhances it. As such, the logic to collate the data should be in its own class IMO, rather than making the liveblog.php even bigger.

As I was looking into this, I then saw that (far too) _many_ of the methods in the various classes seem to be `static`, so my initial desired approach of making the Liveblog_Metadata follow the _Decorator_ pattern (class that accepts an interface which it also implements, so that it wraps the default Liveblog class) would need a re-think (and then I went on holiday!).

I think that merging your branch into mine, and then re-checking mine and merging it here would make sense for _now_, but it doesn't address what I think is some technical debt that is causing a limitation on using good coding practices.

philipjohn:
> it doesn't address what I think is some technical debt that is causing a limitation on using good coding practices

:100: we have a bunch of technical debt to address, and we will :)

PR #513, opened 2018-08-14
Update readme.txt

maevelander:
Adding @philipjohn as a key contributor. Updating link to documentation now that VIP internal docs moved to a password protected area.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -1,5 +1,5 @@
 === Liveblog ===
-Contributors: automattic, nbachiyski, batmoo, johnjamesjacoby
+Contributors: automattic, nbachiyski, batmoo, johnjamesjacoby, philipjohn
 Tags: liveblog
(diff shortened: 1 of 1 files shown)


readme.txt
```diff
@@ -25,7 +25,7 @@ Here’s what makes it special:
  * Your authors can drag-and-drop photos right into the Liveblog area, without needing to navigate to separate browser tabs or windows.
  * There’s no need for a separate site dedicated to liveblogging: *every* post can be a liveblog, even existing ones.
 
-[Check out our in-depth documentation.](http://vip.wordpress.com/documentation/using-the-liveblog-plugin/)
+[Check out our in-depth documentation.](https://github.com/Automattic/liveblog#liveblog)
```
GaryJones: Using `#readme` would also work, and it is then decoupled from the heading wording.

PR #514, opened 2018-08-15
Add an option to disable the static endpoint cache 

rogertheriault:
for users who need fine-grain control over the endpoint base, this allows selective or site-wide disabling of the static cache.

This might be needed for a site that uses more than one domain, for example a different domain for wp-admin vs the front end.

### DIFF
--- a/classes/class-wpcom-liveblog-rest-api.php
+++ b/classes/class-wpcom-liveblog-rest-api.php
@@ -28,7 +28,12 @@ class WPCOM_Liveblog_Rest_Api {
 
 	public static function build_endpoint_base() {
 
-		if ( ! empty( self::$endpoint_base ) ) {
+		/**
+		 * static caching is enabled by default
+		 *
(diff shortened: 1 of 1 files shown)


mjangda:
Should the static cache just be removed? How expensive is it to build the endpoint base?

rogertheriault:
I was thinking it should probably be dropped, but was not keen to break someone's site if they'd set up some elaborate filtering to get around the inability to filter :-) so @rebeccahum suggested just adding a filter that could be used just by affected sites.

I'm not clear why it was there. Static caches are handy but maybe not so handy for something that folks expect to be able to filter. It doesn't look like it saves that much processing.

But removing it entirely is probably something for a future release by @philipjohn

rogertheriault:
> Do we know of any one wanting to do this, or is it just a thing we think people might want?
Only the folks we implemented the filter for in SVN right now. I'm ok with just removing the static cache instead, if it's unnecessary.

Issue #504, opened 2018-07-29
Updates stop when paginating AMP Liveblog


philipjohn:
When viewing an AMP Liveblog, it provides a notification of new entries. These cease as soon as pagination is used.

# Steps to reproduce

1. Create a Liveblog with more than one page
2. Open the AMP version
3. Post a new update via the non-AMP liveblog
4. A notification will appear on the AMP liveblog
5. Click through to page 2 on the AMP liveblog
6. Post a new entry on the non-AMP liveblog
7. You will not see a notification on the AMP liveblog
8. Click back to page 1
9. There will be no new entries or notifications shown until you refresh.

GaryJones:
> 4. A notification will appear on the AMP liveblog


![screenshot 2018-08-06 20 04 06](https://user-images.githubusercontent.com/88371/43735951-ec395fe8-99b3-11e8-9bb6-61d84689808b.png)

> 9. There will be no new entries or notifications shown until you refresh.

I couldn't even get it to show by refreshing - I had to remove the `liveblog_last` querystring parameter.

@philipjohn Question: What should happen if I was viewing page 2 in AMP, the notification did appear, and I clicked on it? Should I be sent back to page 1? Some other behaviour?


Aside: Deletion of entries does NOT appear to produce a notification in the AMP Liveblog.

GaryJones:
Interestingly, the Preview of the Ajax response _DOES_ include the notification.

![screenshot 2018-08-06 20 14 29](https://user-images.githubusercontent.com/88371/43736401-573f2ee8-99b5-11e8-952f-790dc79bb5da.png)

So it seems that the view isn't getting updated for the user (by whatever mechanism is supposed to do that), even though the data is there to do so.

philipjohn:
> @philipjohn Question: What should happen if I was viewing page 2 in AMP, the notification did appear, and I clicked on it? Should I be sent back to page 1? Some other behaviour?

My expectation would be to get returned to page 1 with the new entries loaded.

> Aside: Deletion of entries does NOT appear to produce a notification in the AMP Liveblog.

I think we'd want those to silently disappear - what do you think?

> So it seems that the view isn't getting updated for the user (by whatever mechanism is supposed to do that), even though the data is there to do so.

Yep looks like it - presumably the state isn't changing properly. I'll have to defer to @jasonagnew on this one as I'm not familiar enough with this part.

jasonagnew:
Hey @philipjohn @GaryJones @maevelander 

[...]

PR #519, opened 2018-08-28
AMP Pagination Fix

jasonagnew:
This PR resolves the issues discussed in Issue: https://github.com/Automattic/liveblog/issues/504

When pagination is used in AMP, the update button should still be displayed and take the user back to page 1.

### DIFF
--- a/classes/class-wpcom-liveblog-amp.php
+++ b/classes/class-wpcom-liveblog-amp.php
@@ -250,6 +250,11 @@ class WPCOM_Liveblog_AMP {
 
 		$request = self::get_request_data();
 
+		// If AMP Polling request don't restrict content so it knows there is updates are available.
+		if ( self::is_amp_polling() ) {
+			$request->last = false;
+		}
+
 		if ( $request->id ) {
 			$entries  = WPCOM_Liveblog::get_entries_paged( false, false, $request->id );
 			$request  = self::set_request_last_from_entries( $entries, $request );
@@ -354,6 +359,7 @@ class WPCOM_Liveblog_AMP {
 				'page'     => $entries['page'],
 				'pages'    => $entries['pages'],
 				'links'    => self::get_pagination_links( $request, $entries['pages'], $post_id ),
+				'last'     => get_query_var( 'liveblog_last', false ),
 				'settings' => array(
 					'entries_per_page' => WPCOM_Liveblog_Lazyloader::get_number_of_entries(),
 					'refresh_interval' => WPCOM_Liveblog::get_refresh_interval(),
@@ -431,6 +437,7 @@ class WPCOM_Liveblog_AMP {
 
 		$permalink = amp_get_permalink( $post_id );
 
+		$links['base']  = self::build_paged_permalink( $permalink, 1, false );
 		$links['first'] = self::build_paged_permalink( $permalink, 1, $request->last );
 		$links['last']  = self::build_paged_permalink( $permalink, $pages, $request->last );
 
@@ -502,5 +509,18 @@ class WPCOM_Liveblog_AMP {
 	public static function get_template( $name, $variables = array() ) {
 		$template = new WPCOM_Liveblog_AMP_Template();
 		return $template->render( $name, $variables );
+	}
+
+	/**
+	 * Is this an AMP polling request.
+	 *
+	 * @return boolean AMP polling request.
+	 */
+	public static function is_amp_polling() {
+		$amp_latest_update_time = filter_input( INPUT_GET, 'amp_latest_update_time', FILTER_SANITIZE_STRING );
+		if ( ! empty( $amp_latest_update_time ) ) {
+			return true;
+		}
+		return false;
 	}
 }
--- a/templates/amp/feed.php
+++ b/templates/amp/feed.php
@@ -4,6 +4,7 @@ $post_id          = $this->get( 'post_id' );
 $links            = $this->get( 'links' );
 $page             = $this->get( 'page' );
 $pages            = $this->get( 'pages' );
+$last             = $this->get( 'last' );
 $settings         = $this->get( 'settings' );
 $entries_per_page = $settings['entries_per_page'];
 $refresh_interval = $settings['refresh_interval'];
@@ -24,7 +25,11 @@ $social           = $settings['social'];
 
 	<button id="live-list-update-button"
 		update
-		on="tap:amp-live-list-insert-blog.update"
+		<?php if ( $last === false ): ?>
+			on="tap:amp-live-list-insert-blog.update"
+		<?php else: ?>
+			on="tap: AMP.navigateTo(url='<?php echo esc_url( $links->base ); ?>')"
+		<?php endif ?>
 		class="ampstart-btn caps"><?php esc_html_e( 'You have updates' ); ?></button>
 	<div items>
 


classes/class-wpcom-liveblog-amp.php
```diff
@@ -503,4 +510,17 @@ public static function get_template( $name, $variables = array() ) {
...
 		return $template->render( $name, $variables );
 	}
+
+	/**
+	 * Is this an AMP polling request.
+	 *
+	 * @return boolean AMP polling request.
+	 */
+	public static function is_amp_polling() {
+		$amp_latest_update_time = filter_input( INPUT_GET, 'amp_latest_update_time', FILTER_SANITIZE_STRING );
```
GaryJones: While the `filter` extension (that provides the `filter_input()` function) is enabled by default, it can also be disabled. See https://secure.php.net/manual/en/filter.installation.php

IIRC, WordPress core doesn't rely on this for this exact reason.

If this extension is to be a requirement, then it should be added to the `composer.json` `require` section, and also add a note to the READMEs.

GaryJones: The `'amp_latest_update_time'` feels like a magic string - could it be added as a class constant instead?

classes/class-wpcom-liveblog-amp.php
```diff
@@ -503,4 +510,17 @@ public static function get_template( $name, $variables = array() ) {
...
+		if ( ! empty( $amp_latest_update_time ) ) {
```
GaryJones: This conditional and extra return could be simplified:

```php
return ! empty( $amp_latest_update_time );
```

classes/class-wpcom-liveblog-amp.php
jasonagnew: Hey @GaryJones 

Thanks for feedback. As I'm not using the value the ideal way would be:

```
	public static function is_amp_polling() {
		return isset( $_GET[ AMP_UPDATE_QUERY_VAR ] ); // input var okay.
	}
```
But that will result in `WARNING | Processing form data without nonce verification.`

I could add `filter` to `composer.json` but I guess trying to avoid using `filter` may make more sense if its not used anywhere else. The query var is added by the AMP plugin, unfortunately it doesn't appear possible to access it using `get_query_var()`

GaryJones: If the nonce really isn't needed here (false positive), then the following should disable the warning:

```php
public static function is_amp_polling() {
	// phpcs:ignore WordPress.Security.NonceVerification.NoNonceVerification
	return isset( $_GET[ AMP_UPDATE_QUERY_VAR ] );
}
```



> I guess trying to avoid using filter may make more sense if its not used anywhere else.

There are two existing uses of `filter_input()`, both in the root `liveblog.php` file. The predicament would obviously also apply to them, but since this PR wouldn't make the situation worse, I can't argue about it - it should have its own PR to discuss whether it should be a recognised dependency or not :-)

classes/class-wpcom-liveblog-amp.php
```diff
@@ -7,6 +7,11 @@
  */
 class WPCOM_Liveblog_AMP {
 
+	/**
+	 * AMP adss the following query string to requests when polling.
```
GaryJones: Typo: "adss" => adds

classes/class-wpcom-liveblog-amp.php
```diff
@@ -517,10 +522,7 @@ public static function get_template( $name, $variables = array() ) {
 	 * @return boolean AMP polling request.
 	 */
 	public static function is_amp_polling() {
-		$amp_latest_update_time = filter_input( INPUT_GET, 'amp_latest_update_time', FILTER_SANITIZE_STRING );
-		if ( ! empty( $amp_latest_update_time ) ) {
-			return true;
-		}
-		return false;
+		// phpcs:ignore WordPress.Security.NonceVerification.NoNonceVerification
+		return isset( $_GET[ self::AMP_UPDATE_QUERY_VAR ] ); // input var okay.
```
GaryJones: The `// input var okay` comment can be removed. This is from the old WPCS whitelisting, but is now covered by the `// phpcs:ignore ...` on the line above.

jasonagnew: Thanks, good to know. I've sorted that now 👍

classes/class-wpcom-liveblog-amp.php
```diff
@@ -503,4 +515,14 @@ public static function get_template( $name, $variables = array() ) {
 		$template = new WPCOM_Liveblog_AMP_Template();
 		return $template->render( $name, $variables );
 	}
+
+	/**
+	 * Is this an AMP polling request.
+	 *
+	 * @return boolean AMP polling request.
```
GaryJones: Should be `bool`, rather than `boolean`.

templates/amp/feed.php
```diff
@@ -24,7 +25,11 @@
 
 	<button id="live-list-update-button"
 		update
-		on="tap:amp-live-list-insert-blog.update"
+		<?php if ( $last === false ): ?>
+			on="tap:amp-live-list-insert-blog.update"
+		<?php else: ?>
+			on="tap: AMP.navigateTo(url='<?php echo esc_url( $links->base ); ?>')"
```
GaryJones: I don't know React. This has a space after the `tap:`, but the one on line 29 doesn't. Is the inconsistency intentional?

jasonagnew: I think it makes sense to have no space. When checking the example from an AMP doc it had the space so I assumed it was way to do it. However after doing some searching you don't have to include it and it seems more common to not have the space.

GaryJones:
Following the steps at https://github.com/Automattic/liveblog/issues/504#issue-345532248, this does work (the notification appears), though you have to wait on page 1 until the next poll (15 seconds?).

Is there a way for this notification to be shown instantly, since it was already checked for and retrieved once?

jasonagnew:
Hey @GaryJones 

Thanks for the feedback, just wanted to outline the three situations I see and what currently happens.

1) The user is on the first page; javascript based update: http://bigbite.im/2lsd5H
2) The user is on another page; update button redirects them to page 1 where the update is already visible http://bigbite.im/ctWxG6
3) The user is on another page and wishes to continue reading without clicking the update. The reason liveblog keeps your last updated entry is to give you fixed pages. In our version, let's say the user wanted to count from 24 which starts on page 3, all the way to 38 on page 1. If we don't fix the pages then they would count from 24 to 28, Editor publishes an entry 39, User changes to page 2, which would start at 30, not 29, in our version we keep their page structure so this doesn't happen: http://bigbite.im/tSUtGD

My assumption is on number 3 you would want the update button to appear straight away when they land on page 1? If that's the case, the only solution I can see is dropping the poll time right down if we know the user has come through to page 1 with a fixed pages set. We have very little control over how the update feature works as its AMP javascript library that calls the HTML page and compares with what you see to decide if there are updates. Although the main Liveblog plugin is React, the AMP work is not.

Thanks
Jason

GaryJones:
Thanks for taking the time record those videos.

Having understood the problem a little more, I agree that what you've implemented now seems to be the best resolution for the different scenarios. 👍

jasonagnew:
Hey @GaryJones 

Just a heads up, I spoke @philipjohn and he asked me to change it so when moving back to page 1 it won't fix your position so will see the updates right away.

Thanks
Jason

philipjohn:
Thanks Jason, this is :+1:

Issue #476, opened 2018-06-03
Show progress indicator


philipjohn:
When users perform an action such as adding, editing or deleting an entry it would be good to provide some sort of progress indicator so that there is visual feedback that something is happening.

davidsword:
I'll take this one!

PR #527, opened 2018-09-21
Feature/476 show progress indicator

davidsword:
If on a slower network/server, when clicking the 'Publish New Entry', 'Publish Update Entry', or 'Delete' entry buttons, the user receives no feedback, no indication that their click is actually doing anything. This can be confusing or lead to repetitive clicks or miss-clicks.

The following commits solve this problem by adding states during these 'transactions' which the components can conditionally check and show progress indicators by changing text or classNames as needed.

The Delete buttons `isDeleting` boolean state usage is all self contained and handled within EntryContainer instead of using Store, as once the Entry is deleted, the component will be removed from state, unmounting, and no further state change is required.

### DIFF
--- a/src/react/containers/EditorContainer.js
+++ b/src/react/containers/EditorContainer.js
@@ -87,7 +87,7 @@ class EditorContainer extends Component {
   }
 
   publish() {
-    const { updateEntry, entry, entryEditClose, createEntry, isEditing } = this.props;
+    const { updateEntry, entry, createEntry, isEditing } = this.props;
     const { editorState, authors } = this.state;
     const content = this.getContent();
     const authorIds = authors.map(author => author.id);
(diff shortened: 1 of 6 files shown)


src/react/containers/EditorContainer.js
```diff
@@ -87,7 +87,7 @@ class EditorContainer extends Component {
   }
 
   publish() {
-    const { updateEntry, entry, entryEditClose, createEntry, isEditing } = this.props;
```
sboisvert: Why are we removing entryEditClose?

davidsword: This commit explains it: https://github.com/Automattic/liveblog/pull/527/commits/0ef80388b0b6bc9e4faa9771c1562ded6ae8edcd

> This function was used to close the editor immediately after clicking 'Publish Update'. When a user on a slower network/server clicked this, the editor disappeared instantly, however it would take several seconds for the entries text to actually update - there was no visual feedback that their action was processing.
> 
> By not using this function the editor does not instantly get removed. Not being removed right away allows the editor component to check if the `store > user > entries` entry has a true isPublishing state, adding necessary indicators of progress.
> 
> The editor is already set to close by changing the isEditing state to false, on the `UPDATE_ENTRY_SUCCESS` Redux action in the user.js reducer.

davidsword: Ah yes @sboisvert, I see the problem now, I had that in the wrong commit. My mistake.

sboisvert:
I feel like the build failures here are unrelated.

davidsword:
@sboisvert the failing I believe is: https://github.com/Automattic/liveblog/pull/529

GaryJones:
Looks like there's some stylelint items to address, and a merge conflict.

GaryJones:
Closing this PR as it's 6 years old with merge conflicts that can't be automatically resolved (the source fork has been deleted). The underlying feature request is tracked in #476 which remains open. Thank you @davidsword for the detailed implementation work\!

Issue #526, opened 2018-09-13
Review usage of dangerouslySetInnerHTML


nickdaugherty:
@uxcitizen noticed that there are several spots in this plugin using React's `dangerouslySetInnerHTML`.

https://github.com/Automattic/liveblog/search?q=dangerouslysetinnerhtml&unscoped_q=dangerouslysetinnerhtml

How many of those can we get rid of with refactoring?

Obviously some places are showing pre-rendered HTML from the server. In those cases, we should be sanitizing the HTML to strip out any unexpected or disallowed tags and attributes before passing it to `dangerouslySetInnerHTML` to reduce the potential for abuse.

Ideally we have 0 of these calls, but for the ones we can't get away from, we should be sure everything is sanitized as far as possible.

cain:
Seems like all of the WordPress endpoints return HTML, it's not possible without using `dangerouslySetInnerHTML`

But you are right, some sanitizing needs to be done. https://github.com/cure53/DOMPurify could be used.

davidsword:
I'll look into this!

cain:
@davidsword looks really good!

PR #530, opened 2018-09-25
add sanitization to dangerouslySetInnerHTML values. Re: #526

davidsword:
There are several spots in this plugin using Reacts `dangerouslySetInnerHTML` attribute to place content from the server. This can easily lead to cross-site scripting (XSS) attacks.

Before the HTML is dangerously placed, running it through [DOMPurify](https://github.com/cure53/DOMPurify) helps to sanitize out malicious HTML.

`iframe`'s (an easy method for XSS), are allowed for the embedded media, however the domain used must be whitelisted - otherwise the `src` is removed.

The domains for the iframe whitelist reflect the front end "Embed Media" list https://github.com/Automattic/liveblog/blob/master/README.md#embedding-media 

⚠️  This sanitization, by design, removes any `<script>` and/or non-whitelisted `<iframe>` a user has added themselves in an entry, including content within 'HTML Blocks'. When merged into a upcoming release, this should be mentioned in the *Upgrade Notice*.

### DIFF
--- a/src/react/utils/utils.js
+++ b/src/react/utils/utils.js
@@ -1,3 +1,5 @@
+import DOMPurify from 'dompurify';
+
 import moment from './extendedMoment';
 
 /* eslint-disable no-param-reassign */
@@ -199,4 +201,36 @@ export const getScrollToId = (entries, key) => {
   }
 
   return `id_${entries[0].id}`;
+};
+
+/**
+ * Sanitize HTML for output. Help prevent XSS attacks.
+ *
+ * @link https://www.npmjs.com/package/dompurify
+ * @param  {string} dirty HTML to sanitize
+ * @return {string} sanitized HTML
+ */
+export const sanitizeHTML = (dirty) => {
+  // Whitelist iframes for the plugins 'embded media' feature.
+  const iframeWhitelist = [
+    'www.hulu.com',
+    'player.hulu.com',
+    'open.spotify.com',
+    'player.vimeo.com',
+    'www.youtube.com',
+    // Instagram and Twitter don't use iframes.
+  ];
+  const regex = RegExp(`^(https:|)//(${iframeWhitelist.join('|')})/`, 'im');
+  DOMPurify.addHook('afterSanitizeAttributes', (node) => {
+    if (node.tagName === 'IFRAME') {
+      const iframeSrc = node.getAttribute('src');
+      if (iframeSrc && !iframeSrc.match(regex)) {
+        node.removeAttribute('src');
+      }
+    }
+  });
+  const clean = DOMPurify.sanitize(dirty, {
+    ADD_TAGS: ['iframe'],
+  });
+  return clean;
 };
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "brace": "^0.11.1",
     "core-js": "^2.5.1",
     "custom-event-polyfill": "^0.3.0",
+    "dompurify": "^1.0.8",
     "draft-convert": "^2.0.0",
     "draft-js": "^0.10.3",
     "draft-js-export-html": "^1.2.0",
--- a/src/react/Editor/convertToHTML.js
+++ b/src/react/Editor/convertToHTML.js
@@ -4,6 +4,8 @@
 import React from 'react';
 import { convertToHTML } from 'draft-convert';
 
+import { sanitizeHTML } from '../utils/utils';
+
 export default contentState =>
   convertToHTML({
     styleToHTML: () => {},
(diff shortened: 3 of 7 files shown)


cain:
@davidsword Looks like you need to build the assets :)

sboisvert:
This looks great!

I suspect @cain is correct in needing to rebuild the compiled JS to get it working.

One thing that came to mind is, what happens if someone has an extra domain they want whitelisted. For example to their own site or to some other service (that they trust)? Is there a way we could have a filter in the plugin where they could add domains?

src/react/utils/utils.js
```diff
@@ -200,3 +202,35 @@ export const getScrollToId = (entries, key) => {
...
+ */
+export const sanitizeHTML = (dirty) => {
+  // Whitelist iframes for the plugins 'embded media' feature.
+  const iframeWhitelist = [
+    'www.hulu.com',
+    'player.hulu.com',
+    'open.spotify.com',
+    'player.vimeo.com',
+    'www.youtube.com',
+    // Instagram and Twitter don't use iframes.
```
mjangda: Does this mean we will strip out Twitter and Instagram embeds?

mjangda: Also, how was this list of domains generated?

sboisvert:
Chatted with @mjangda on this. 
A few great questions that came from it:

1) What is our motivation behind this? 
Is the content that's being fed to the `dangerouslySetInnerHTML` something that isn't being validated by WP? If it isn't, why not? If it is, what are we going to do about things that we validate in WP and accept and then filter out in JS? If it's impossible for something to come to the JS without having been thru PHP validation, is the increase in security worth breaking current backwards compatibility?

2) What will this break?
If we're introducing a breaking change, should we first notify users in a version? (For example the next release could lists all the things that would be removed and link to documentation on how to validate / whitelist it properly)
Is there a way we could look over the most common things this would strip and then figure out if those should be whitelisted by default?

src/react/utils/utils.js
```diff
@@ -200,3 +202,35 @@ export const getScrollToId = (entries, key) => {
...
+
+/**
+ * Sanitize HTML for output. Help prevent XSS attacks.
+ *
+ * @link https://www.npmjs.com/package/dompurify
+ * @param  {string} dirty HTML to sanitize
+ * @return {string} sanitized HTML
+ */
+export const sanitizeHTML = (dirty) => {
+  // Whitelist iframes for the plugins 'embded media' feature.
```
paulschreiber: Typo in `embed`

davidsword:
Thank you all for the input and questions. I will answer and look into this more this week.

davidsword:
The custom endpoints on the REST API that are feeding `dangerouslySetInnerHTML` are not sanitized. If database was compromised, an XSS attack could output from the REST API and the React app would display it. 

Since XSS prevention is needed and no sanitization is occurring (besides on new entry submissions), wether we do it with PHP at the custom endpoints, or with JS using DOMPurify, it seems breaking changes are inevitable. In reviewing this again, I now believe both locations should be sanitized using the same filter-able white/blacklist. This way we're not serving stored XSS and we're also escaping as late as possible. 

I'm not sure *why* things are in this current state, perhaps a broader history lessons needs to happen. It looks like the old PHP version of the plugin did have `wp_kses()` and [an extra whitelist of allowed tags](https://github.com/davidsword/liveblog/commit/9368bda5e3bb3b639c05fcc399b5a5e28e91c694#diff-33d85e4895defcbe898196482a22be8cR8) before output - however the current REST/React setup does not use it.

As for _what_ DOMPurify breaks (as the PR currently is), here are the implicit blacklist for [tags](https://github.com/cure53/DOMPurify/wiki/Default-TAGs-ATTRIBUTEs-whitelist-&-blacklist#html-tags) & [attrs](https://github.com/cure53/DOMPurify/wiki/Default-TAGs-ATTRIBUTEs-whitelist-&-blacklist#html-attributes) _(whitelist: [tags](https://github.com/cure53/DOMPurify/blob/master/src/tags.js) & [attrs](https://github.com/cure53/DOMPurify/blob/master/src/attrs.js))_. This PR only varies from the defaults by whitelisting `iframe` with our own function to validate `src` values. I can see the need for a discussion around what the defaults should be.

& thanks for bringing up the questions. This PR lacked plugin documentation for *what* is default white/blacklisted, and Upgrade Notice details on breaking change, and so on. More info needed.

As well as having the white/blacklist be filterable, and documentation on that. I will add that in as this goes along.

src/react/utils/utils.js
davidsword: This list of domains was generated taking liveblogs advertised media embed features, adding them to an entry, and seeing what WordPress/oembed gave back. In writing that, I realize I need to take a closer look into these domains (ie: subdomain changes for different regions or languages) and each services documentation to ensure its a single primary domain for everything.

& as pointed out by @sboisvert, a filter to manage this whitelist would be needed.

GaryJones:
@davidsword Any interest in fixing the merge conflicts and finishing this one off?

Issue #525, opened 2018-09-13
LiveBlog at the top of the article


cmjaimet:
I would like to have the option to place LiveBlogs at the top of posts rather than below them. We serve mobile stories adaptively and insert sidebar widgets into the content. With the liveblog at the bottom of the content, it gets pushed down by ads and other widgets.

The code that places the liveblog block is in /liveblog.php L1138: 
```
public static function add_liveblog_to_content( $content ) { 
	... 
	return $content . wp_kses_post( $liveblog_output ); 
}
```

I would prefer: 
```
if ( true === $show_liveblog_at_top ) { 
	return wp_kses_post( $liveblog_output ) . $content; 
} else { 
	return $content . wp_kses_post( $liveblog_output ); 
}
```

I would be happy to fork the repo and write this code if the change is approved.

This would involve adding either a global or a post-level setting (or both) to set/get $show_liveblog_at_top in post meta or options.

For the time being I am adding a filter to my functions.php file. This is a fragile hack and relies on particular HTML existing in the LiveBlog output, but it works for now. 

```
function pm_move_liveblog_to_top( $content ) { 
	$liveblog_marker = '<div id="wpcom-liveblog-container"'; 
	$start_position = strpos( $content, $liveblog_marker ); 
	if ( false !== $start_position ) { 
		$pieces = explode( $liveblog_marker, $content ); 
		$content = $liveblog_marker . $pieces[1] . $pieces[0]; 
	} 
	return $content; 
} 
add_filter( 'the_content', 'pm_move_liveblog_to_top', 21 );
```
Charles

cmjaimet:
Thanks this would solve our problem nicely.

deepakrohillas:
public static function add_liveblog_to_content( $content ) {

			// We don't want to add the liveblog to other loops
			// on the same page
			if ( ! self::is_viewing_liveblog_post() ) {
				return $content;
			}

			$liveblog_output = '<div id="wpcom-liveblog-container" class="' . self::$post_id . '"></div>';

			$liveblog_output = apply_filters( 'liveblog_add_to_content', $liveblog_output, $content, self::$post_id );
			if(false === apply_filters( 'liveblog_output_at_bottom' ,true ) ) {
				return  wp_kses_post( $liveblog_output ) . $content;
			} else {
				return  $content.wp_kses_post( $liveblog_output );
			}
		}

Filter :  
add_filter( 'liveblog_output_at_bottom', function (){
	return false;
} );

PR #531, opened 2018-09-26
Add a filter to modify Liveblog output location

alexiskulash:
Fixes: #525

Some users would prefer having Liveblog output at the top of their posts rather than the bottom. This filter introduces the simple functionality to choose whether the Liveblog output should be above content or below it (default).

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -1208,7 +1208,17 @@ if ( ! class_exists( 'WPCOM_Liveblog' ) ) :
 			$liveblog_output = '<div id="wpcom-liveblog-container" class="' . self::$post_id . '"></div>';
 
 			$liveblog_output = apply_filters( 'liveblog_add_to_content', $liveblog_output, $content, self::$post_id );
-			
+
+			/**
+			 * Filters whether the liveblog output should appear at the top of the post content.
+			 *
+			 * By default, the liveblog container is appended after the post content.
+			 * Return true to prepend the liveblog before the post content instead.
+			 *
+			 * @since 1.10.0
+			 *
+			 * @param bool $at_top Whether to display the liveblog at the top. Default false.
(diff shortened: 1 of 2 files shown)


liveblog.php
```diff
@@ -1169,6 +1169,10 @@ public static function add_liveblog_to_content( $content ) {
 			$liveblog_output = '<div id="wpcom-liveblog-container" class="' . self::$post_id . '"></div>';
 
 			$liveblog_output = apply_filters( 'liveblog_add_to_content', $liveblog_output, $content, self::$post_id );
+			
+			if ( false === apply_filters( 'liveblog_output_at_bottom', true ) ) {
```
sboisvert: I'm not sure why but to me it feels more obvious to have the filter that changes something to be "liveblog_output_at_top" and have it default to false. I'm having a hard time to vocalize it but it seems like a negation that doesn't need to be there. Like the question is liveblog_output_at_bottom and then having it to true will will then output it to the top. It works, but it "feels" strange to me.

alexiskulash: I think that makes sense; it kind of feels like a double negative when folks want to actually have output at the top, whereas a default false for 'liveblog_output_at_top' feels more straightforward.

GaryJones:
The failure here appears to be related to the version number that was in `packages.json`. It was changed [here](https://github.com/Automattic/liveblog/commit/63df4d0042146accbb845c0e9b93c62cf6283afa#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L4). Perhaps it was because `1.9.0-RC2` not a published release?

I suspect that re-starting the build will make this pass.

GaryJones:
Restarting the build produced the same result, but I think rebasing it onto the latest `master` may help.

PR #539, opened 2018-10-08
Add support for other post types

shantanu2704:
Submitting a new PR because the changes from [this PR](https://github.com/Automattic/liveblog/pull/342) were lost.

1. Added a static variable for supported post types that can be filtered with the `liveblog_modify_supported_post_types` filter in the `init()` method.
2. Addressed the feedback for `page` post type in the previous PR. 
3. Modified `get_liveblog_state()` and `is_viewing_liveblog_post()` to accommodate additional post types.

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -59,6 +59,7 @@ if ( ! class_exists( 'WPCOM_Liveblog' ) ) :
 		public static $auto_archive_days       = null;
 		public static $auto_archive_expiry_key = 'liveblog_autoarchive_expiry_date';
 		public static $latest_timestamp        = false;
+		public static $supported_post_types    = array();
 
 
 		/** Load Methods **********************************************************/
@@ -250,7 +251,11 @@ if ( ! class_exists( 'WPCOM_Liveblog' ) ) :
 			 * Add liveblog support to the 'post' post type. This is done here so
 			 * we can possibly introduce this to other post types later.
 			 */
-			add_post_type_support( 'post', self::KEY );
+			$post_types = array( 'post' );
+			self::$supported_post_types = apply_filters( 'liveblog_modify_supported_post_types', $post_types );
+			foreach ( self::$supported_post_types as $post_type ) {
+				add_post_type_support( $post_type, self::key );
+			}
 
 			/**
 			 * Apply a Filter to Setup our Auto Archive Days.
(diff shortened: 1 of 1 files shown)


liveblog.php
```diff
@@ -250,7 +251,11 @@ public static function init() {
 			 * Add liveblog support to the 'post' post type. This is done here so
 			 * we can possibly introduce this to other post types later.
 			 */
-			add_post_type_support( 'post', self::KEY );
+			$post_types = array( 'post' );
+			self::$supported_post_types = apply_filters( 'liveblog_modify_supported_post_types', $post_types );
```
GaryJones: This filter should be [documented](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters).

paulschreiber: I would rename it to `liveblog_supported_post_types`

liveblog.php
```diff
@@ -252,7 +252,16 @@ public static function init() {
...
 			 */
 			$post_types = array( 'post' );
-			self::$supported_post_types = apply_filters( 'liveblog_modify_supported_post_types', $post_types );
+
+			/**
+			 * Filters post types supported by liveblog.
+			 *
+			 * @since 2.0
+			 *
+			 * @param array An array of supported post types.
```
GaryJones: The type is already specified as an array, so the description can just be `Supported post types.`

liveblog.php
```diff
@@ -252,7 +252,16 @@ public static function init() {
 			 * we can possibly introduce this to other post types later.
 			 */
 			$post_types = array( 'post' );
-			self::$supported_post_types = apply_filters( 'liveblog_modify_supported_post_types', $post_types );
+
+			/**
+			 * Filters post types supported by liveblog.
```
GaryJones: There is a plan to possibly change the name of the plugin, so maybe change `liveblog` here to be `this plugin`?

shantanu2704: Are we changing the name in this release?
If not, I think this should stay as `liveblog` as it'll be easier to change the prefix with a search-replace when the name is finalised.

shantanu2704: what do you think?

GaryJones: The prefix on the filter is unlikely to change, unless it was aiming to break BC. 

In this case, it was just the name in the filter description (line 257), that I was referring to.

Not a biggie - there are likely other instances of "liveblog" in other DocBlocks as well that would need changing.

shantanu2704: ah, got it! thanks!

PR #486, opened 2018-06-07
PHPCS fixes for 1.9.1-dev

paulschreiber:

PR #541, opened 2018-10-16
Reinstate PHPCS and update VIP sniffs

### LINKED ISSUE #540: Fix PHPCS sniffs on PRs
> The command "if [[ "$SNIFF" == "1" ]]; then $PHPCS_DIR/scripts/phpcs -p -s -v -n . --standard="WordPress-VIP" --extensions=php; fi" exited with 0.

It looks like PHPCS isn't running as expected via Travis on Pull Requests. Rather than failing, the Travis tests are passing, masking the fact that there's even an issue.

We need to:

- [ ] Ensure PHPCS runs reliably via Travis
- [ ] Include the WordPress Coding Standards
- [ ] Include the `WordPress-VIP-Go` ruleset from [VIP Coding Standards](https://github.com/Automattic/VIP-Coding-Standards)

philipjohn:
This PR re-instates the PHPCS sniffs via Travis. It also includes the new VIP Coding Standards repo and runs the newer `WordPress-VIP-Go` standard.

Fixes #540

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ script:
     # -n flag: Do not print warnings. (shortcut for --warning-severity=0)
     # --standard: Use WordPress as the standard.
     # --extensions: Only sniff PHP files.
-    - if [[ "$SNIFF" == "1" ]]; then $PHPCS_DIR/bin/phpcs -p -s -v -n . --standard="WordPress-VIP-Go" fi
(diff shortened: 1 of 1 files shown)


philipjohn:
The only failures here are because the PHPCS check (now added back) is flagging some issues. We don't need to address those in this PR. They should (hopefully) be addressed in #486

PR #547, opened 2018-10-17
update print_liveblog_metadata() to pass parameters to get_liveblog_metadata()

paulschreiber:
In `fix/phpcs`, we discovered that `get_liveblog_metadata()`'s method signature was missing parameters (even though they were listed in the headerdoc).

I added the parameters to the method definition and to one of the calls, but missed the `print_liveblog_metadata` call.

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -1847,7 +1847,9 @@ if ( ! class_exists( 'WPCOM_Liveblog' ) ) :
 				return;
 			}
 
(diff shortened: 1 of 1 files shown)


paulschreiber:
@philipjohn

paulschreiber:
The fix in `9dd494d97b908f187d74969f690a1dac5c3ba20b` is incorrect, as it passes a post ID, not a post object. The ID is then dereferenced, ($post->ID), which will fail.

PR #548, opened 2018-10-18
Update packages to fix sub-dependency security issues

GaryJones:
`npm audit` showed that there were four paths to the `hoek` package, which has a known security issue: https://www.npmjs.com/advisories/566

This change is the result of running `npm update node-gyp --depth 2`, as advised by the `npm audit` results. `node-gyp` and all four paths are related to the top-level `node-sass` package.

`npm audit` now reports no vulnerabilities.

Note: this does not address https://github.com/Automattic/liveblog/network/alert/package-lock.json/cryptiles/open

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "liveblog",
-  "version": "1.8.2",
+  "version": "1.9.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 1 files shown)


GaryJones:
Travis is failing on an unrelated PHPCS issue.

paulschreiber:
This should be done via yarn. See https://github.com/Automattic/liveblog/pull/550

PR #550, opened 2018-10-19
Update node modules

paulschreiber:
* remove `package-lock.json`, as `yarn` is used
* use tabs to indent `package.json`
* update yarn.lock


cain:
good stuff

GaryJones:
> * as `yarn` is used

Where is it documented that Yarn is used exclusively?

Other than the lock files, there is no difference seen in the repo.

The history of the files suggests they've both had about the same amount of updates. The yarn lock file was introduced [here](https://github.com/Automattic/liveblog/pull/322/files) but there was no discussion on that switch happening.

Yarn _was_ useful, for performance and lock files, before NPM got its act together and added a whole load of new features (including lock files, security audits, `npx` usage, CLI autocompletion, as good or better performance than Yarn and all the features Yarn has etc.)

Npm _comes_ with node, but Yarn is another bit of software a contributor needs to install to be able to contribute, if this PR goes ahead. When the first-party tool does the job, I don't see a reason to exclusively require a different tool.

Like HHVM with PHP, Yarn was Facebook's internally-developed tool that brought new features vs npm, and like PHP, npm saw massive improvements in response to the point where the Facebook-created tooling isn't needed by the general community.

References (top few results after searching for `yarn vs npm 2018`) - all of them recommend, to one degree or another, using npm over yarn:
 - [Yarn vs npm - which Node package manager to use in 2018?](https://blog.risingstack.com/yarn-vs-npm-node-js-package-managers/)
 - [Why Are You Still Using Yarn in 2018?](https://iamturns.com/yarn-vs-npm-2018/)
 - [To Yarn and Back (to npm) Again](https://mixmax.com/blog/to-yarn-and-back-again-npm/)
 - [2018 Yarn vs. NPM](https://blog.alexwendland.com/2018/yarn-vs-npm/)

I did a quick performance test, with clean caches. Yarn took 24 seconds:

screenshot 2018-10-19 09 54 16


NPM took 21 seconds, AND gave me a warning about security issues (which is what #548 fixes):

screenshot 2018-10-19 09 53 26

I agree that maintaining two lock files is a good way to get them out of sync, but if anything, I'd say that the yarn lock file should be the one to be removed.

GaryJones:
> * use tabs to indent `package.json`

+1 for this, though whitespace formatting should probably be in its own PR, and not included with the potential remove of support for any node package manager.

GaryJones:
> It seems like an oversight that package-lock.json was not removed when yarn was added

Maybe, but the history of both files suggests that folks have generally been keeping them both up to date ever since - yarn isn't used exclusively.

> • yarn is faster

I've shown that it's not with the screenshots, albeit with a sample of one size on one machine. I definitely used to be faster, but the HHVM used to be faster than PHP, and for the latest versions, that's no longer true either.

Even if there was a potential speed benefit (there isn't, on average), it's a separate tool that is unnecessary, when those using Node already have npm.

> • Travis is configured to use yarn. Local development should match that.

That's easily changeable. Removing a dependency is a good thing. Even the comment before the `yarn` call says about using `npm` scripts.

> • `yarn install` generates a warning:

That's a good thing, and another reason to get rid of one of the lock files. While yarn advises about the presence of a particularly-named file that _may_ cause some workflow frustrations later on, `npm` warns of security issues, and has a way to automatically fix them. I know which I'd prefer.

paulschreiber:
I've closed this and filed separate PRs:
* Use tabs for indenting package.json https://github.com/Automattic/liveblog/pull/552
* Update yarn, remove package-lock.json: https://github.com/Automattic/liveblog/pull/553
or
* Update nom, remove yarn.lock, update Travis config https://github.com/Automattic/liveblog/pull/554

PR #554, opened 2018-10-19
Update npm packages, remove yarn.lock

paulschreiber:


GaryJones:
Noting for reviewers, that the `package-lock.json` hasn't actually [changed](https://github.com/Automattic/liveblog/pull/554/commits/390a656537671e9f62e82541c7f148977cddc3dd) much.

I checked out this PR, and did `npm i`. NPM still reported the 4 `npm audit` issues which #548 addresses, so while this PR moves us forward in terms of not having two package lock files (:+1:), it doesn't fix the security issues which triggered this discussion.

GaryJones:
I also note that after my `npm i`, my `package-lock.json` tries to remove the `"optional": "true"` that this PR adds. I'm using node 8.11.3, and npm 6.4.1. I suspect that @paulschreiber is using a different version.

Worth considering (perhaps as a different PR), of minimum Node / NPM / engine versions, to avoid this in the future?

paulschreiber:
I ran `npm update node-gyp --depth 2 ` to update the packages. I am also using node 8.11.3 and npm 6.4.1.

GaryJones:
> I ran `npm update node-gyp --depth 2` to update the packages.

So now this is duplicating what was already done in #548.

Why not just revert the changes to package-lock.json, and just keep this PR focused on removing the Yarn / the Yarn lock file?

jasonagnew:
It's worth noting we built the project with Yarn, I know @paulschreiber you added the `package-lock.json` about 3 months ago but the `yarn.lock` as been around since October 2017. I'd personally prefer to stay with Yarn as all previous releases have been built using it. The only advantage to using NPM would be that fact other big WordPress projects (for example Gutenberg) use it so people may be more familiar. This is probably a documentation issue as there's nothing explaining how to compile assets.

paulschreiber:
I am happy to go with yarn. Gary didn’t like it. I made two pull requests, covering both possibilities.

GaryJones:
@jasonagnew

> The only advantage to using NPM

I'd [disagree](https://github.com/Automattic/liveblog/pull/550#issuecomment-431296564) that its use on Gutenberg et al is the only reason.

jasonagnew:
@GaryJones No problem. I would however recommend we don't switch when doing the smaller releases, so for example 1.9.2. Although in theory they are interchangeable its probably not worth the risk of an issue being brought into a release on compile. For 2.0 it could be a good chance if you want to switch as its going to have some big changes.

GaryJones:
> I would however recommend we don't switch when doing the smaller releases, so for example 1.9.2.

I'm fine with that waiting until a major release.

philipjohn:
Thanks for the discussion folks, I resolved the conflicts and merged #556

PR #558, opened 2018-10-19
Send the correct data to get_liveblog_metadata()

### LINKED ISSUE #557: WPCOM_Liveblog::print_liveblog_metadata() passing post ID instead of WP_Post object
In 1.9.1 I'm seeing this error:
```
Trying to get property 'ID' of non-object in /var/www/wp-content/plugins/liveblog/liveblog.php on line 1792
```

I'll submit a PR to change [liveblog.php#L1853](https://github.com/Automattic/liveblog/blob/e5dc95289f7cd98d709e58eed9b75c86c8534305/liveblog.php#L1853) to send `get_post()` instead of `get_the_ID()`.

brettshumaker:
Switched from `get_the_ID()` to `get_post()` in `WPCOM_Liveblog::print_liveblog_metadata()`.

`WPCOM_Liveblog::print_liveblog_metadata()` was sending a post ID to `WPCOM_Liveblog::get_liveblog_metadata()` instead of a post object causing a `Trying to get property 'ID' of non-object` error.

Fixes #557

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -1850,7 +1850,7 @@ if ( ! class_exists( 'WPCOM_Liveblog' ) ) :
 				return;
 			}
 
-			$metadata = self::get_liveblog_metadata( array(), get_the_ID() );
+			$metadata = self::get_liveblog_metadata( array(), get_post() );
 			if ( empty( $metadata ) ) {
 				return;
 			}


paulschreiber:
See also #547

Issue #417, opened 2018-04-26
Author list is refetched on each keypress


paulschreiber:
The author list is refetched on each keypress. This is very slow. It should be fetched once and cached, and the matching should be done client-side.

philipjohn:
Fixed in #568

paulschreiber:
The "fix" reduces the time between fetches. Only one fetch is needed per page load. No need to re-fetch the author list.

PR #568, opened 2018-10-22
Add debounce to author list async box

jasonagnew:
Added a debounce to the author list box on editor to resolve issue #417 - underscores is used for the debounce, in an ideal world we would use RxJS debounce but doesn't appear possible within React Select.

Debounce is set to 500ms but theres a filter to change that `liveblog_author_list_debounce_time`

Please note I did not include complied assets so merging is cleaner.

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -1070,6 +1070,13 @@ if ( ! class_exists( 'WPCOM_Liveblog' ) ) :
 						'class_alert'                  => __( 'type-alert', 'liveblog' ),
 						'class_key'                    => __( 'type-key', 'liveblog' ),
 
+						/**
+						 * Filters the Author list debounce time, defaults to 500ms.
+						 *
+						 * @since 1.9.2
+						 *
+						 * @param int $time Author list debounce time.
+						 */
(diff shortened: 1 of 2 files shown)


liveblog.php
```diff
@@ -1068,6 +1069,8 @@ public static function enqueue_scripts() {
 						'class_term_prefix'            => __( 'term-', 'liveblog' ),
 						'class_alert'                  => __( 'type-alert', 'liveblog' ),
 						'class_key'                    => __( 'type-key', 'liveblog' ),
+
+						'author_list_debounce_time'    => apply_filters( 'liveblog_author_list_debounce_time', self::AUTHOR_LIST_DEBOUNCE_TIME ),
```
GaryJones: Filter should be [documented](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters).

(I know most of the others aren't yet, but it's got to start somewhere, and new code is a good a place as any!)

philipjohn:
Tested this patch by loading a liveblog and watching the network requests in the browser as I typed in the author field. Before this patch each keypress resulted in a request to the `author` endpoint. After the patch is applied there is a visible delay and substantial reduction in the number or requests.

Issue #560, opened 2018-10-22
Shortcode quotes decode issue


dhanendrannewscorp:
We have recently updated the Liveblog version to 1.9 and after that we are facing some issues with the shortcode rendering.

we found that the liveblog send the data as HTML encoded to the server. So we need to decode them before processing it. 
https://github.com/Automattic/liveblog/blob/master/classes/class-wpcom-liveblog-rest-api.php#L654

For example, we are trying to add a video embed into the entry box and it sends the data like

```

<p>[video width=&quot;1280&quot; height=&quot;720&quot; mp4=&quot;https://wwwexample.com/wp-content/uploads/2018/10/slow_motion_rain.mp4&quot;][/video]</p>

```

So we need to add `html_entity_decode()` [here](https://github.com/Automattic/liveblog/blob/master/classes/class-wpcom-liveblog-rest-api.php#L654)

philipjohn:
Possibly related to #418 but needs investigating

philipjohn:
Fixed by #570

Issue #418, opened 2018-04-27
Numeric entities are chopped up with hashtags


combs:
We are seeing some users pasting in rich text that comes with numeric entities, like `&#x27;`

The liveblog plugin is turning them into hashtags 
https://www.washingtonian.com/2018/04/27/chat-ann-limpert-friday-6/

![image](https://user-images.githubusercontent.com/6174018/39376585-33eddf8c-4a20-11e8-854d-6758fa501577.png)

Can the hashtag parsing be improved to prevent this?

cain:
Currently using https://github.com/Automattic/liveblog/commit/712b2119c76007debd92e8725662880b0ac9eaf7 
for a quick fix. Would love to know when this gets fixed in master! :)

philipjohn:
Fixed by #570

PR #570, opened 2018-10-23
Fix encoding issues which affect shortcodes and hashtags

jasonagnew:
This PR resolves issues #560 and #418 

We now decode content when it hits the server using `html_entity_decode()` which fixes both shortcodes and hashtags. The hashtag regex has also been updated to handle any situation where hex code does find its way into the content.

### DIFF
--- a/classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
+++ b/classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
@@ -39,6 +39,9 @@ class WPCOM_Liveblog_Entry_Extend_Feature_Hashtags extends WPCOM_Liveblog_Entry_
 		// the generated hashtag class.
 		$this->class_prefix = apply_filters( 'liveblog_hashtag_class', $this->class_prefix );
 
+		// Set a better regex for hashtags to allow for hex values in content.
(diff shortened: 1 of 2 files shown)


philipjohn:
Just noting that test failures are unrelated to code in this PR

classes/class-wpcom-liveblog-entry-extend-feature-hashtags.php
```diff
@@ -39,6 +39,9 @@ public function load() {
 		// the generated hashtag class.
 		$this->class_prefix = apply_filters( 'liveblog_hashtag_class', $this->class_prefix );
 
+		// Set a better regex for hashtags to allow for hex values in content.
+		$this->set_regex( '~(?:(?<!\S)|>?)(?:(?<!&|&amp;))((?:' . implode( '|', $this->get_prefixes() ) . ')([0-9_\-\p{L}]*))~um' );
```
GaryJones: Where has this regex come from? Where's the unit tests to show it working? Where's the documentation / breakdown, like [here](https://github.com/WordPress/WordPress/blob/da7a80d67fea29c2badfc538bfc01c8a585f0cbe/wp-includes/shortcodes.php#L243).

I've no doubt that it works, but for the sake of reducing technical debt moving forward, I'd like to see this PR be rounded out more.

Issue #475, opened 2018-06-03
Entries can be empty


philipjohn:
The editor allows you to post an empty entry, which probably isn't desirable. We should either prevent publishing or have some sort of confirmation.

no-sws:
@philipjohn I had a couple ideas of how to approach this that I wanted to throw out there.

Re: testing for an empty entry, one idea I had was to use JavaScript's `DOMParser` class along with the HTML string value provided by the editor, e.g.,

```javascript
// Example editor value
const editorValue = '<p>Hello, World <img src="/abc.png"></p>';

const parser = new DOMParser();
const doc = parser.parseFromString(editorValue, 'text/html');

if (doc.images.length < 1 && '' === doc.body.textContent.trim()) {
  // Empty...
} else {
  // Not empty...
}
```

My concern with that approach is security (i.e., does creating an in-memory doc from the editor value create any vulnerabilities, such as potential script execution?) and browser compatibility (although a polyfill could be an option). It may also be omitting some cases, like video elements.

As an alternative (or complement), we could maybe just use a regular expression(s) to determine if the content was empty (e.g., strip tags and entities and trim) and check for any image or video content.

Re: the user interface, I figured we could either use a simple `window.confirm` call (if the entry was empty) or add one or more state vars to the `EditorContainer` container component to keep track of whether publishing had been attempted (and the entry was empty). Depending on how we wanted to approach it, we could show a simple message or provide a "publish override" that would allow the user to publish empty content.

Look forward to your feedback. Thanks!

cain:
Whats the opinion on `window.confirm`. I was thinking of using it at #436

philipjohn:
> does creating an in-memory doc from the editor value create any vulnerabilities, such as potential script execution?

I'm honestly not sure. How might that be exploitable?

Would using DOMParser add significant extra code to the plugin? What extra memory usage is that likely to add on the browser and how might that impact performance, especially on mobile?

> As an alternative (or complement), we could maybe just use a regular expression(s) to determine if the content was empty (e.g., strip tags and entities and trim) and check for any image or video content.

How likely is it that we'll be able to craft a good enough regex to cover everything? What kind of memory usage is running that regex likely to create?

> I figured we could either use a simple window.confirm

I like simple :) We can iterate on it in future if there's a call for something more sophisticated.

no-sws:
[...]

PR #572, opened 2018-10-23
Stop editors from publishing empty entries.

jasonagnew:
This is a simple way to stop editors publishing empty entries, resolves issue #475

I didn't include complied assets to make merging cleaner

### DIFF
--- a/src/react/containers/EditorContainer.js
+++ b/src/react/containers/EditorContainer.js
@@ -93,6 +93,17 @@ class EditorContainer extends Component {
     const authorIds = authors.map(author => author.id);
     const author = authorIds.length > 0 ? authorIds[0] : false;
     const contributors = authorIds.length > 1 ? authorIds.slice(1, authorIds.length) : false;
+    const htmlregex = /<(img|picture|video|audio|canvas|svg|iframe|embed) ?.*>/;
+
+    // We don't want an editor publishing empty entries
+    // So we must check if there is any text within the editor
+    // If we fail to find text then we should check for a valid
+    // list of html elements, mainly visual for example images.
+    if (!editorState.getCurrentContent().getPlainText().trim()) {
+      if (htmlregex.exec(convertToHTML(editorState.getCurrentContent())) === null) {
+        return;
+      }
+    }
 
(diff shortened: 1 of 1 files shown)


src/react/containers/EditorContainer.js
```diff
@@ -93,6 +93,17 @@ class EditorContainer extends Component {
     const authorIds = authors.map(author => author.id);
     const author = authorIds.length > 0 ? authorIds[0] : false;
     const contributors = authorIds.length > 1 ? authorIds.slice(1, authorIds.length) : false;
+    const htmlregex = /<(img|picture|video|audio|canvas|svg|iframe|embed) ?.*>/;
+
+    // We don't want an editor publishing empty entries
```
GaryJones: This functionality seems like it should be within its own function, so that the consuming code here would be as simple as (pseudo-code):

```js
// Don't allow publishing of empty entries.
if ( editorIsEmpty( editorState ) ) {
    return;
}
```

The logic is then abstracted into something like (again, pseudo-code):

```js
/**
 * Check if the content is empty.
 *
 * Look for text within the editor, the most commen case. If that's not found, look for markup indicating media / non-text content.
 *
 * @since 1.9.2
 * (Plus whatever the JS equivalent of `@param` and `@return` tags are.)
 */
editorIsEmpty( editor ) {
	const htmlregex = /<(img|picture|video|audio|canvas|svg|iframe|embed) ?.*>/;
    const content   = editor.getCurrentContent();
	
    if ( content.getPlainText().trim() ) {
        // There is text content.
        return false;
    }

    if ( htmlregex.exec(convertToHTML(content)) === null) {
        // There is some non-text content.
        return false;
    }

    return true;
}
```

(The code would be potentially more re-usable and testable, if you just pass in `editorState.getCurrentContent()`, rather than the instance of the `editorState`.)

cain:
Good stuff!

philipjohn:
Tested and this fixes the issue so while I like the idea of extracting out the check a little, this does the job nicely for now. I'll happily look at a PR for an enhancement to this though :)

PR #590, opened 2018-11-01
Wrong parameter passed to get_liveblog_metadata()

### LINKED ISSUE #589: Wrong parameter passed to get_liveblog_metadata()
**Issue 1:**
In [`print_liveblog_metadata()`](https://github.com/Automattic/liveblog/blob/master/liveblog.php#L1859) function, there is a small mistake in the [`get_liveblog_metadata()`](https://github.com/Automattic/liveblog/blob/master/liveblog.php#L1866) function call.

Instead of passing [`WP_Post`](https://github.com/Automattic/liveblog/blob/master/liveblog.php#L1797) object, post id been passed in.

**Issue 2:**
In the [DocBlock](https://github.com/Automattic/liveblog/blob/master/liveblog.php#L868) of `entries_for_json()`, param type should be mentioned.

dhanendrannewscorp:
Fixes #589

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -1863,7 +1863,7 @@ if ( ! class_exists( 'WPCOM_Liveblog' ) ) :
 				return;
 			}
 
-			$metadata = self::get_liveblog_metadata( array(), get_the_ID() );
+			$metadata = self::get_liveblog_metadata( array(), get_post() );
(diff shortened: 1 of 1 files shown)


GaryJones:
Fixed in #558.

PR #597, opened 2018-11-06
remove get_fields_for_render()

### LINKED ISSUE #596: Remove get_fields_for_render()
This is dead code.

paulschreiber:
Remove get_fields_for_render(), which appears to be unused. Fixes #596.


GaryJones:
Noting, that at least one client is using this.

A method call should not have been removed in a minor release (1.9.5).

PR #607, opened 2018-12-06
Pass post object, not just the ID.

philipjohn:
Fixes a PHP Notice caused by `get_liveblog_metadat()` expecting a full post object but only receiving a post ID.

> Notice: Trying to get property 'ID' of non-object in /wp-content/plugins/liveblog/liveblog.php on line 1809


GaryJones:
Handled in https://github.com/Automattic/liveblog/pull/558.

PR #610, opened 2018-12-17
Do not load liveblog entry if entry ID is not set

Ritesh-patel:


GaryJones:
Under what conditions might an entry not have an entry.id?

PR #659, opened 2020-05-29
Fix the travis builds

david-binda:
Start the mysql service, update the bin/install-wp-tests.sh script, and fix linting issues


GaryJones:
Travis CI no longer in use.

PR #663, opened 2020-07-03
AMP: Fix too high refresh interval

philipjohn:
The refresh interval can be set by the site owner to any value. On AMP liveblogs this is used for the `data-poll-interval` in the `amp-live-list` but that value has to be greater than 15. https://amp.dev/documentation/examples/components/amp-live-list/

This change ensures that AMP Liveblogs have a minimum interval of 15.


classes/class-wpcom-liveblog-amp.php
```diff
@@ -332,7 +332,7 @@ public static function build_entries_feed( $entries, $request, $post_id ) {
 				'last'     => get_query_var( 'liveblog_last', false ),
 				'settings' => array(
 					'entries_per_page' => WPCOM_Liveblog_Lazyloader::get_number_of_entries(),
-					'refresh_interval' => WPCOM_Liveblog::get_refresh_interval(),
+					'refresh_interval' => min( 15, WPCOM_Liveblog::get_refresh_interval() ),
```
paulschreiber, line 335: Maybe put this in a constant.

PR #668, opened 2020-12-03
Add jsonpFunction to prevent a conflict

srtfisher:
**Problem:**
When loading the plugin with other webpack installations active, a Javascript conflict arises and throws a fatal console error. The live blog fails to load on the page.

**Solution:**
Adds a `jsonpFunction` option to the webpack output to prevent a conflict with other webpack installations (brings liveblog into its own little webpack bubble).

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -18,6 +18,7 @@ const webpackConfig = {
   },
 
   output: {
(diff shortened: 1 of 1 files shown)


GaryJones:
I believe this is solved with https://github.com/Automattic/liveblog/pull/660.

PR #674, opened 2021-06-05
setup github actions

trepmal:
- removes travis config (where builds are currently failing; travis-ci.org will also be shut down June 15th)
 - moves tests, phpcs, linting to github actions


.github/workflows/lint.yml
```diff
@@ -0,0 +1,38 @@
...
+    # Prevent the "push" build from running when there are only irrelevant changes.
+    push:
+        paths-ignore:
+            - "**.md"
+    pull_request:
+    # Allow manually triggering the workflow.
+    workflow_dispatch:
+
+jobs:
+    test:
```
GaryJones, line 14: Probably want a different name for this job.

.github/workflows/cs-lint.yml
```diff
@@ -0,0 +1,70 @@
+name: CS & Lint
```
GaryJones, line 1:
```suggestion
name: PHP CS & Lint
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,38 @@
+name: Run linting
```
GaryJones, line 1:
```suggestion
name: Run JS linting
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,38 @@
...
+    push:
+        paths-ignore:
+            - "**.md"
+    pull_request:
+    # Allow manually triggering the workflow.
+    workflow_dispatch:
+
+jobs:
+    test:
+        name: Run lint
```
GaryJones, line 15:
```suggestion
        name: Run JS linting
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,38 @@
...
+
+            - name: Install node
+              uses: actions/setup-node@v2
+              with:
+                  node-version: "14"
+
+            - name: NPM install
+              run: npm install
+
+            - name: Run lint-scripts
```
GaryJones, line 33:
```suggestion
            - name: Run eslint on React files
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,38 @@
...
+              run: npm run lint-scripts
+
+            - name: Run lint-styles
```
GaryJones, line 36:
```suggestion
            - name: Run styleslint on SCSS files
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,38 @@
...
+              if: ${{ always() }}
```
GaryJones, line 37: What does this line add?

GaryJones (changes requested):
I think the new linting for JS files could be a little more descriptive to identify what it's doing, and how that differs from the "regular" linting for PHP.

.github/workflows/cs-lint.yml
GaryJones: Alternatively, leave this as "CS & Lint", and rename the lint.yml to "CS & Lint" as well, and then leave the specifics to the job name. See how the actions are reported to see how best to name/group/identify workflows/jobs/steps:

Screenshot 2021-06-23 at 09 48 03

GaryJones:
Tests are looking for a `prepare-ci` Composer script, but only `prepare` exists.

.github/workflows/lint.yml
trepmal: ensures that `run lint-styles` will run even if the preceding step fails.

.github/workflows/lint.yml
trepmal: Honestly hadn't occurred to me `test` was variable. Is there a preferred convention here or just the slug'd name, e.g. `run_js_linting`?

GaryJones: I'm not sure the [job_id](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_id) is actually exposed anywhere, as the "name" is shown instead, so as long as it's unique for the `jobs`, then whatever you want is fine; in this case though, calling it `test` and not actually running any tests didn't make sense to me.

.github/workflows/lint.yml
GaryJones: Would a [`continue-on-error`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) on the previous step be a better approach, as then each and every potential future step doesn't need to include the `if` condition.

.phpcs.xml.dist
```diff
@@ -0,0 +1,70 @@
...
+	<description>Custom ruleset for liveblog plugin.</description>
+
+	<!-- For help in understanding this file: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml -->
+	<!-- For help in using PHPCS: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Usage -->
+
+	<!-- What to scan -->
+	<file>.</file>
+	<!-- Ignoring Files and Folders:
+		https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-files-and-folders -->
+	<exclude-pattern>/vendor/</exclude-pattern>
```
GaryJones, line 12: The existing https://github.com/Automattic/liveblog/blob/master/phpcs.xml.dist (which you'll want to remove) also has some extra exclusions that would be worth reviewing to see if they make sense to include here.

.phpcs.xml.dist
```diff
@@ -0,0 +1,70 @@
...
+	<config name="testVersion" value="5.6-"/>
+
+	<!-- Rules: WordPress Coding Standards - see
+		https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards -->
+	<!-- WordPress-Extra includes WordPress-Core -->
+	<rule ref="WordPress-Extra"/>
+	<rule ref="WordPress-Docs"/>
+	<!-- For help in understanding these custom sniff properties:
+		https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/wiki/Customizable-sniff-properties -->
+	<config name="minimum_supported_wp_version" value="5.5"/>
```
GaryJones, line 42: The readme.txt [currently](https://github.com/Automattic/liveblog/blob/f433d63f96342f03fcd921e9956a01b79f11786b/readme.txt#L4) has this as min WP 4.4.

Do you think this plugin counts as a wider-use plugin that should be kept as a wider-supported min WP?

.phpcs.xml.dist
```diff
@@ -0,0 +1,70 @@
...
+
+	<!-- Rules: WordPress VIP - see
+		https://github.com/Automattic/VIP-Coding-Standards -->
+	<rule ref="WordPress-VIP-Go"/>
+
+	<rule ref="WordPress.NamingConventions.PrefixAllGlobals">
+		<properties>
+			<property name="prefixes" type="array">
+				<element value="msm"/>
```
GaryJones, line 51: Nope :D

trepmal: curses!

.github/workflows/lint.yml
trepmal: `continue-on-error` leads to a green checkmark which I think would be misleading https://github.com/Automattic/liveblog/runs/2906734901?check_suite_focus=true

vs the if-always, which will go red while also giving us the results of both steps https://github.com/Automattic/liveblog/pull/674/checks?check_run_id=2906654064

.phpcs.xml.dist
trepmal: This is definitely a plugin to refine the testing on. I've just thrown 4.4 into the matrix as well (but not every version since).

.github/workflows/lint.yml
GaryJones: Huh - it's even [documented](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions) as that. Makes sense.

GaryJones (approved):
Approving what's here, but there may need to be follow-ups to get tests passing.

PR #675, opened 2021-09-09
Implement auto-deployment to the WordPress.org svn repo

philipjohn:
Implements [10up's Github Action](https://github.com/marketplace/actions/wordpress-plugin-deploy) to automatically push new tags to the WordPress.org SVN repo.

The `.distignore` removes anything not necessary for development, or stuff that doesn't make sense (like Markdown files) from SVN. It runs `npm run build` to build the assets before committing. It also moves WordPress.org-specific assets to the `.wordpress-org` directory which the deploy action will put into an assets folder in SVN.


GaryJones (approved):
Seems to be `screenshot-3` missing from the list of moved files, but otherwise looks good.

philipjohn:
Yeah, looks like there just isn't a screenshot-3 for some reason..

PR #676, opened 2021-09-16
FIX: Liveblog scheduled cron event auto_archive_check_hook does not get run

speedyprem:
There will not have any impact on the application, but It schedules the event for the previous midnight. 

This fix will change the schedule to one of the future dates from today midnight to tomorrow.

### DIFF
--- a/classes/class-wpcom-liveblog-cron.php
+++ b/classes/class-wpcom-liveblog-cron.php
@@ -24,7 +24,7 @@ class WPCOM_Liveblog_Cron {
 	private static function configure_autoarchive_cron() {
 
 		if ( ! wp_next_scheduled( 'auto_archive_check_hook' ) ) {
-			wp_schedule_event( strtotime( 'today midnight' ), 'daily', 'auto_archive_check_hook' );
+			wp_schedule_event( strtotime( 'tomorrow midnight' ), 'daily', 'auto_archive_check_hook' );
 		}
 
 		add_action( 'auto_archive_check_hook', array( __CLASS__, 'execute_auto_archive_housekeeping' ) );


philipjohn:
Thanks for this @speedyprem! Can you talk more about the issue you're seeing with this event not firing? I'm not aware of any reports of this issue as yet.

speedyprem:
> Thanks for this @speedyprem! Can you talk more about the issue you're seeing with this event not firing? I'm not aware of any reports of this issue as yet.

@philipjohn It has cropped up in multisite because of the sites are in multiple timezone like UK, USA, and Irish. In this case the event `auto_archive_check_hook` gets scheduled for the previous midnight, so doesn't get run automatically for US sites.

philipjohn:
Hmm okay, that's interesting. Are you seeing Liveblogs remain active when they should have been auto-archived, or was it a backup of cron events that made you aware of the problem?

speedyprem:
> Hmm okay, that's interesting. Are you seeing Liveblogs remain active when they should have been auto-archived, or was it a backup of cron events that made you aware of the problem?

Yes, Liveblogs remain active.

philipjohn:
Are you able to outline your set up for me please? E.g. what sites are in your multisite, what timezone are they on, is Liveblog network active, and any other information that may be relevant? I want to try and replicate the issue to confirm the bug.

speedyprem:
> Are you able to outline your set up for me please? E.g. what sites are in your multisite, what timezone are they on, is Liveblog network active, and any other information that may be relevant? I want to try and replicate the issue to confirm the bug.

> what sites are in your multisite
We have 6 sites with 2 timezone and all are news sites.

> what timezone are they on
1. UK timezone
2. US timezone

Yes, Liveblog is network active for all sites.

PR #682, opened 2023-06-07
Liveblog embed.redditmedia.com javascript file returning 404

deepakrohillas:
Currently in liveblog plugin we are using  wrong reddit media js file ie. https://embed.redditmedia.com/widgets/platform.js?ver=1.9.7

https://embed.redditmedia.com/widgets/platform.js is redirecting on https://embed.reddit.com/widgets.js, which is impacting performance as well as Page speed.
After pushing found if we are adding version then  js file does not work , version based sdk going 404 not found, we don't needs to add version variable in sdk js file

Instagram js also changed https://platform.instagram.com/en_US/embeds.js redirecting on https://www.instagram.com/embed.js
 
reddit
404-reddit404

### DIFF
--- a/classes/class-wpcom-liveblog-entry-embed-sdks.php
+++ b/classes/class-wpcom-liveblog-entry-embed-sdks.php
@@ -40,7 +40,9 @@ class WPCOM_Liveblog_Entry_Embed_SDKs {
 
 		foreach ( self::$sdks as $name => $url ) {
 			if ( 'reddit' === $name ) {
-				wp_enqueue_script( $name, esc_url( $url ), array(), null, false );
+				// Do not attach `?ver=` for reddit script due to redirects
+				$version = 'reddit' === $name ? null : WPCOM_Liveblog::VERSION;
+				wp_enqueue_script( $name, esc_url( $url ), array(), $version, false );
 			} else {
 				wp_enqueue_script( $name, esc_url( $url ), array(), WPCOM_Liveblog::VERSION, false );	
 			}
--- a/classes/class-wpcom-liveblog-entry-embed-sdks.php
+++ b/classes/class-wpcom-liveblog-entry-embed-sdks.php
@@ -39,7 +39,11 @@ class WPCOM_Liveblog_Entry_Embed_SDKs {
 		}
 
 		foreach ( self::$sdks as $name => $url ) {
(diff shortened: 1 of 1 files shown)


classes/class-wpcom-liveblog-entry-embed-sdks.php
```diff
@@ -39,7 +39,7 @@ public static function enqueue() {
 		}
 
 		foreach ( self::$sdks as $name => $url ) {
-			wp_enqueue_script( $name, esc_url( $url ), array(), WPCOM_Liveblog::VERSION, false );
+			wp_enqueue_script( $name, esc_url( $url ), array(), null, false );
```
rebeccahum, line 42: Can we leave it as-is, please? This should go out with a new release of Liveblog.

deepakrohillas: problem is if version will add in reddit js file like https://embed.reddit.com/widgets.js?ver=123  it will not work 
then again it will generate 404 error , that's why i have removed or we can do ,we can add a conditional statement for this only in case of reddit version will be null and for other's version will add.

deepakrohillas: @rebeccahum 👍 Please check now

classes/class-wpcom-liveblog-entry-embed-sdks.php
```diff
@@ -39,7 +39,11 @@ public static function enqueue() {
 		}
 
 		foreach ( self::$sdks as $name => $url ) {
-			wp_enqueue_script( $name, esc_url( $url ), array(), WPCOM_Liveblog::VERSION, false );
+			if ( 'reddit' === $name ) {
+				wp_enqueue_script( $name, esc_url( $url ), array(), null, false );
```
rebeccahum, line 43: For readability, is it possible to leave a comment for why we are doing it this way and maybe something like this?

```suggestion
                                 $version = 'reddit' === $name ? null : WPCOM_Liveblog::VERSION;
				wp_enqueue_script( $name, esc_url( $url ), array(), null, false );
```

trepmal (approved):
Looks good to me. 

No ETA on timing of merge and release at this time.

PR #693, opened 2024-07-11
Fix: string interpolation issue on the `WPCOM_Liveblog_Entry_Extend_Feature_Commands` class

renatonascalves:
I noticed this while testing a client site with PHP 8.2

```
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in /var/www/site/wp-content/plugins/liveblog/classes/class-wpcom-liveblog-entry-extend-feature-commands.php on line 243
```

This is usually harmless, but we use `convertDeprecationsToExceptions=true` in our unit tests, which essentially fails because of this PHP deprecation.

The fix works for any PHP version, so it should be safe for a minor version bump. :)


mjangda (approved):

renatonascalves:
@mjangda any update on merging this?

PR #698, opened 2024-12-11
Update class-wpcom-liveblog-entry-extend-feature-commands.php

jberculo:
${type} is deprecated. Changed to {$type}


GaryJones (approved):
This fix is correct. The `${var}` syntax for string interpolation in double-quoted strings is deprecated as of PHP 8.2 and will be removed in PHP 9.0. The proper syntax is `{$var}`.

I verified this is the only instance of this deprecated pattern in the codebase that needs fixing (other `${...}` occurrences are in single-quoted strings intended as JavaScript templates).

Thank you for the contribution!

PR #699, opened 2024-12-14
Update WP_CLI::line() to WP_CLI::log()

raamdev:
The docs for `WP_CLI::line()` recommend using `WP_CLI::log()`:

> WP_CLI::log() is typically recommended and WP_CLI::line() is included only for historical compat.

https://make.wordpress.org/cli/handbook/references/internal-api/wp-cli-line/

This PR updates references to `WP_CLI::line()` to use `WP_CLI::log()` instead.


GaryJones (approved):

GaryJones:
The CI workflows are out of date, but this change is safe anyway, so we can tackle that separately.

Issue #700, opened 2024-12-17
Outdated packages with vulnerabilities


peter-stewart-thg:
We've been using this package for a number of years, but are currently seeing issues flagged up by dependabot.
Many of the NPM packages are outdated and have vulnerabilities.

Currently we're seeing 100 vulnerabilities:
- **Critical: 13**
- **High: 54**
- Moderate: 24
- Low: 9

I've started auditing the package to see if I can easily resolve these vulnerabilities, however, currently I'm seeing a number of deprecated packages used, so some if not all of these will need to be replaced.
There are also a number of packages that are no longer maintained.

We would like to continue using the package, however, these security vulnerabilities make that not an option for us at the moment.

I aim to complete my audit of the packages today, and will update this issue with my findings.
I will then see how feasible it is to resolve the vulnerabilities.

Do you plan to continue maintaining this package?
Is anyone else working on updating these packages?

If the package is going to be maintained, 
I'm happy to help resolve these issues and get it back up and running.
Also, if there is someone who knows the package well, it would be very helpful to be able to discuss the packages that are currently used.

peter-stewart-thg:
I have been going through the packages this morning, and here is my initial analysis:


# NPM Packages

## Dev Dependencies

- **autoprefixer**  
  NPM: https://www.npmjs.com/package/autoprefixer  
  GitHub: https://github.com/postcss/autoprefixer  
  Last updated: 3 years ago  
  Current version: ^7.1.6  
  Latest version: 10.4.14  
  Vulnerabilities: None reported  
  **Deprecated**: No  
  Alternative: N/A  

- **babel-core**  
  NPM: https://www.npmjs.com/package/babel-core  
  GitHub: https://github.com/babel/babel  
  Last updated: 7 years ago  
  Current version: ^6.25.0  
  Latest version: 6.26.3  
  Vulnerabilities: 2 Critical  
  **Deprecated**: Yes  
  Alternative: https://www.npmjs.com/package/@babel/core  

- **babel-eslint**  
  NPM: https://www.npmjs.com/package/babel-eslint  
  GitHub: https://github.com/babel/babel-eslint  
  Last updated: 5 years ago  
  Current version: ^7.2.3  
  Latest version: 10.1.0  
  Vulnerabilities: 2 Critical  
  **Deprecated**: Yes  
  Alternative: https://www.npmjs.com/package/@babel/eslint-parser  

- **babel-jest**  
  NPM: https://www.npmjs.com/package/babel-jest  
[...]

PR #701, opened 2024-12-18
fix: Updated dependencies to resolve vulnerabilities

peter-stewart-thg:
There were a number of outdated packages,
as well as many that had either been depricated or moved.

Some packages are no longer needed, as they were polyfilling
functionality that now exists later versions of JS or React.

I have replaced the use of 'react-loadable' with
Lazy and Suspense from React.

I've also migrated to the new Babel Mono-repo package structure.

I've updated the webpack config to resolve some issues with changes
in the config parameter as well and switching out
'extract-text-webpack-plugin' for 'mini-css-extract-plugin'
and 'eslint-loader' for 'eslint-webpack-plugin'

The issues related to this PR can be seen here: https://github.com/Automattic/liveblog/issues/700


liveblog.php
```diff
@@ -1542,6 +1542,7 @@ public static function handle_query_vars_for_post_filtering( $query ) {
 		 * @return bool
 		 */
 		public static function current_user_can_edit_liveblog() {
+			$user = wp_get_current_user();
```
GaryJones, line 1545: Where is this new assigned variable being used?

PR #764, opened 2025-12-19
check template part variable names against allow list

psorensen:
Iteration of #760. Introduces stricter check for allowed variables.

### DIFF
--- a/liveblog.php
+++ b/liveblog.php
@@ -1305,7 +1305,7 @@ if ( ! class_exists( 'WPCOM_Liveblog' ) ) :
 
 			foreach ( $template_variables as $key => $value ) {
 				// Only allow variables from the allowlist
-				if ( is_string( $key ) && in_array( $key, $allowed_variables, true ) ) {
+				if ( in_array( $key, $allowed_variables, true ) ) {
 					${$key} = $value;
 				}
 			}


liveblog.php
```diff
@@ -1272,9 +1272,40 @@ private static function get_all_entry_output() {
...
+				'class',
+				'new_label',
+				'new_button',
+			);
+
 			foreach ( $template_variables as $key => $value ) {
-				// Validate key is a valid PHP variable name
-				if ( is_string( $key ) && preg_match( '/^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$/', $key ) ) {
+				// Only allow variables from the allowlist
+				if ( is_string( $key ) && in_array( $key, $allowed_variables, true ) ) {
```
paulschreiber: If you want to allow developers to add additional variables, you could wrap `$allowed_variables` in a filter.

psorensen: I actually confused myself with how this works. The variables are coming from the plugin and are available for developers to use in their custom templates. Seems like the filter may not even be necessary.

GaryJones (commented):
@psorensen's clarification is correct - theme developers create custom templates that *receive* these variables, they don't *define* new ones. The plugin controls what variables are passed via `get_template_part()` calls.

So the allowlist as-is is the right approach:
- **Secure by default** - only known plugin variables are extracted
- **Complete** - I verified all `get_template_part()` calls in the codebase and the list covers them all
- **No filter needed** - themes consume variables, they don't add them

The only scenario where a filter would be useful is if a plugin/theme hooks into `liveblog_before_insert_entry` or similar and calls `get_template_part()` with custom variables - but that seems like an edge case that doesn't warrant the added complexity.

**Minor suggestion:** The `is_string($key)` check is redundant since `in_array()` with strict mode will only match strings against the string allowlist, but it doesn't hurt to keep it for defensive coding.

PR #785, opened 2025-12-29
fix: strip images from key event titles

### LINKED ISSUE #464: Key events title uses anything

cnaples79:
## Summary

Fixes the key event title formatting to properly strip HTML tags, preventing images and other content from leaking into key event headings.

## Problem

The previous code used `wp_strip_all_tags()` with what appeared to be an allowed tags parameter:

```php
wp_strip_all_tags( $content, '<strong><em><span><img>' );
```

However, `wp_strip_all_tags()` doesn't support an allowlist - its second parameter is `$remove_breaks` (boolean). The string was coerced to `true`, so all tags were stripped anyway, but not intentionally.

## Solution

Use `wp_strip_all_tags()` without parameters to produce clean plain text titles:

```php
wp_strip_all_tags( $content );
```

Key event titles in the widget are now guaranteed to be plain text.

## Test plan

- [ ] Create a liveblog entry starting with an image, mark as key event
- [ ] Verify the key events widget shows text content, not the image

Fixes #464

🤖 Updated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/classes/class-wpcom-liveblog-entry-key-events.php
+++ b/classes/class-wpcom-liveblog-entry-key-events.php
@@ -347,8 +347,8 @@ class WPCOM_Liveblog_Entry_Key_Events {
 		// Grab the first sentence of the content.
 		$content = preg_replace( '/(.*?[?!.](?=\s|$)).*/', '\\1', $content );
 
-		// Strip it of all non-accepted tags.
-		$content = wp_strip_all_tags( $content, '<strong></strong><em></em><span></span><img>' );
+		// Strip all HTML tags for plain text title.
+		$content = wp_strip_all_tags( $content );
 
 		return $content;
 	}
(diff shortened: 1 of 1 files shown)


GaryJones (commented):
Thanks for tackling #464! The fix intention is correct, but there's an underlying bug in the existing code that should be addressed:

**`wp_strip_all_tags()` doesn't support an allowlist** - its second parameter is just a boolean for `$remove_breaks`:

```php
wp_strip_all_tags( string $text, bool $remove_breaks = false ): string
```

The current code passes a string like `'<strong><em><span>'` which PHP coerces to `true`, so it just strips all tags AND removes line breaks. The formatting tags were never actually preserved.

**Suggested fix** - use PHP's native `strip_tags()` which does support an allowlist:

```php
// Before (broken)
$content = wp_strip_all_tags( $content, '<strong></strong><em></em><span></span><img>' );

// After (correct)
$content = strip_tags( $content, '<strong><em><span>' );
```

Or if we actually want plain text with no formatting in key event titles:

```php
$content = wp_strip_all_tags( $content );
```

Which approach would you prefer?

PR #829, opened 2026-01-26
Fix fatal error in /entries endpoint

Ritesh-patel:
In the /entries endpoint it fetches all the entries including deleted comment. The delete entries don't have `$comment->comment_date_gmt` set and hence the `DateTimeImmutable` approach introduced in this commit https://github.com/Automattic/liveblog/commit/1dbcf5edd22601f2bf06a7f6e5f77f84ef742f1a doesn't work and throws a fatal error for /entries endpoint.

This fix checks whether `$comment->comment_date_gmt` is not empty and only then uses `DateTimeImmutable` otherwise falls back to use `mysql2date`.

### DIFF
--- a/classes/class-wpcom-liveblog-entry.php
+++ b/classes/class-wpcom-liveblog-entry.php
@@ -182,7 +182,7 @@ class WPCOM_Liveblog_Entry {
 		$comment = get_comment( $comment_id );
 
 		// For Unix timestamp format, use DateTimeImmutable to avoid timezone issues with mysql2date.
-		if ( 'U' === $d || 'G' === $d ) {
+		if ( ( 'U' === $d || 'G' === $d ) && ! empty( $comment->comment_date_gmt ) ) {
 			$datetime = DateTimeImmutable::createFromFormat( 'Y-m-d H:i:s', $comment->comment_date_gmt, new DateTimeZone( 'UTC' ) );
 			return $datetime->getTimestamp();
 		}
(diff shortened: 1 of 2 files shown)


GaryJones (commented):
Thanks for catching this issue, Ritesh!

I have a concern about the fix: when `comment_date_gmt` is empty, the code falls through to `mysql2date()` which will also return incorrect results (likely `false` or an invalid timestamp) for an empty string.

A more robust fix would be to check the return value of `createFromFormat()` rather than pre-checking the input:

```php
if ( 'U' === $d || 'G' === $d ) {
    $datetime = DateTimeImmutable::createFromFormat( 'Y-m-d H:i:s', $comment->comment_date_gmt, new DateTimeZone( 'UTC' ) );
    if ( false !== $datetime ) {
        return $datetime->getTimestamp();
    }
    // Fall through to mysql2date for edge cases
}
```

This handles any malformed date, not just empty ones.

Also worth considering: should deleted entries (which apparently have no `comment_date_gmt`) be filtered out earlier in the query? That might be the root cause to address.

For reference, the 2.x branch handles this in `CommentEntryRepository::hydrate_entry()` with an explicit fallback:

```php
$created_at = DateTimeImmutable::createFromFormat(
    'Y-m-d H:i:s',
    $comment->comment_date_gmt,
    new DateTimeZone( 'UTC' )
);

// Fallback if date parsing fails.
if ( false === $created_at ) {
    $created_at = new DateTimeImmutable( 'now', new DateTimeZone( 'UTC' ) );
}
```

PR #856, opened 2026-04-17
Add French language

dupontbenoit:


dupontbenoit:
The file is translated but I cannot manage to make it work on my server when uploading the file.

GaryJones:
Thanks for the contribution!

Can you please add them to https://translate.wordpress.org/projects/wp-plugins/liveblog/ ? This is where all translations live.

dupontbenoit:
I have added it to the stable project on the plugin page. https://translate.wordpress.org/projects/wp-plugins/liveblog/stable/fr/default/

Do you want me to add it to trunk?

Also, do you have an idea why the (french) translation file is working inside the wordpress admin but not inside the JavasScript editor?

GaryJones:
> Do you want me to add it to trunk?

The next patch release will be this week. After that's done and pushed, we can see if any strings in trunk are different to default.

> Also, do you have an idea why the (french) translation file is working inside the wordpress admin but not inside the JavasScript editor?

Possibly a bug - not had chance to investigate though. Input welcome on what you can find!

GaryJones:
Since the translation files aren't needed directly in this PR, I'll close this out, but continued discussion is welcome.

PR #901, opened 2026-06-02
Disable the deprecated lazyload plugin early on init

GaryJones:
## Summary

While re-reviewing #874 before merge I traced the deprecated-plugin handling that the architecture work had just relocated, and found it had been quietly ineffective for far longer than that.

When the obsolete standalone *Lazyload Liveblog Entries* plugin is active, the bootstrapper is meant to unhook its `init` callback and show an admin notice prompting removal. Both jobs were wired to `template_redirect`, which is the wrong hook for either. `template_redirect` is a front-end-only hook, so `is_admin()` is always false within it and the admin notice could never appear. It also fires *after* `init`, by which point the deprecated plugin's own `init` callback has already run, so unhooking it there is too late to matter. The recent refactor only moved this code; the notice and the unhooking have been dead for much longer.

The fix registers the handler on `init` at priority 0. `init` fires in both admin and front-end contexts, which restores the admin notice, and priority 0 runs before the deprecated plugin's default-priority callback, so the unhooking now takes effect. The lazyload configuration itself stays on `template_redirect`, where it correctly waits for the liveblog query state to be resolved.

To support testing, the handler becomes a public method wired as a named `init` callback rather than an inline closure. That matches every other hook callback in this class (`register_post_type_support`, `register_routes`, and so on) and lets the tests identify it in `$wp_filter`.

## Test plan

- [x] New integration tests assert the handler is hooked on `init` at priority 0 and that invoking it removes the deprecated plugin's `init` callback.
- [x] Lint, PHPCS, the 190 unit tests, and the full 267-test integration suite all pass locally.

### DIFF
--- a/src/php/Infrastructure/WordPress/PluginBootstrapper.php
+++ b/src/php/Infrastructure/WordPress/PluginBootstrapper.php
@@ -425,10 +425,18 @@ final class PluginBootstrapper {
 	 * @return void
 	 */
 	private function init_lazyload(): void {
+		// Disable the deprecated standalone Lazyload Liveblog Entries plugin early
+		// on init, before its own (default-priority) callback fires. This cannot
+		// sit on template_redirect: that hook is front-end only, so is_admin() is
+		// always false there (the admin notice never shows) and it runs after init
+		// (too late to unhook the deprecated plugin).
+		add_action( 'init', array( $this, 'handle_deprecated_lazyload_plugin' ), 0 );
+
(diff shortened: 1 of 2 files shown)


GaryJones:
Superseded by #902, which removes the deprecated standalone *Lazyload Liveblog Entries* plugin shim entirely rather than fixing its notice. As discussed, the plugin has been deprecated since 2015 and is no longer in the wp.org directory, so a major 2.x release is the right place to drop the decade-old back-compat code instead of repairing it. Closing in favour of #902.

PR #904, opened 2026-06-02
Make the admin-ajax state change verify the nonce the client sends

GaryJones:
## Summary

The admin-ajax fallback that toggles liveblog state (used when the REST API is unavailable) verified the nonce under the key `_ajax_nonce` against the action `liveblog_admin_nonce`. Nothing in the plugin produces either of those: `AssetManager` localises, and `src/admin/index.js` sends, the `NONCE_KEY` (`_wpnonce`) request value carrying a `NONCE_ACTION` (`wp_rest`) nonce — the very same nonce the REST path uses. So `wp_verify_nonce()` ran against an empty string and the wrong action on every request, and the handler always returned “Invalid nonce”.

This fails *closed* — the fallback simply never worked, so it is a correctness bug rather than a vulnerability — but it means the non-REST path was dead. It surfaced in the 2.x security review alongside the issues fixed in #903 and was deliberately deferred to here.

The fix verifies the nonce the client actually sends. Rather than read a single named key inside the hook closure (which is where the contract drifted and which could not be tested), the logic now lives in `AdminController::process_state_change()`: a method that takes the request data as an argument, picks `NONCE_KEY` out of it, verifies it against `NONCE_ACTION`, re-checks `edit_post` on the target post, and returns the new meta box markup or a `WP_Error`. `handle_state_change_ajax()` becomes a thin wrapper that forwards the whole request and renders the JSON response. Because the wrapper forwards everything and the tested method chooses the key, the class of bug — glue reading a key the client never sends — can't recur.

## Test plan

Three integration tests on `process_state_change()`:

- a `NONCE_KEY`/`NONCE_ACTION` nonce (what the client sends) is accepted and the state is applied;
- a missing/invalid nonce is rejected and the state is unchanged;
- a valid nonce from a non-owner is still rejected by the post-scoped capability check.

The first was verified to fail against the old `_ajax_nonce`/`liveblog_admin_nonce` contract and pass with the fix. `composer test:integration` — green (277 tests); `phpcs` clean.

### DIFF
--- a/src/php/Infrastructure/WordPress/AdminController.php
+++ b/src/php/Infrastructure/WordPress/AdminController.php
@@ -11,6 +11,7 @@ namespace Automattic\Liveblog\Infrastructure\WordPress;
 
 use Automattic\Liveblog\Application\Config\LiveblogConfiguration;
 use Automattic\Liveblog\Application\Aggregate\LiveblogPost;
+use WP_Error;
 use WP_Post;
 use WP_Query;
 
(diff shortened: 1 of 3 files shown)


GaryJones:
Closing in favour of removing the dead admin-ajax fallback entirely. The endpoint this fixed is unreachable through the bundled client: `USE_REST_API` is a hardcoded `const true` and `can_use_rest_api()` only fails below WP 4.4 (the plugin's own minimum), so `use_rest_api()` is always true and the metabox JS never takes the admin-ajax branch. Rather than secure a path nothing uses, a follow-up will drop the REST-vs-admin-ajax duality and go REST-only.

PR #959, opened 2026-08-10
fix: deduplicate pending polling entries by ID

lstamellos:
## Summary

Prevent cached polling buckets from re-queuing `new` entries whose logical IDs the client already knows about.

The polling start timestamp is intentionally rounded to the refresh-interval bucket for cache efficiency. This can make the requested timestamp range overlap entries already delivered to the client. When new entries are not auto-rendered, a duplicate can enter the pending polling queue and produce a false `1 new entry available` nag.

## Root cause

The start-timestamp bucketing introduced in `99da1326` assumes duplicate polling entries are harmless because entry application is deduplicated by ID.

That is true for rendered API state, but the pending polling queue is separate. A cached `type: "new"` entry can therefore be queued again even though the client already knows that logical entry ID.

There is also a pagination/race variant: rendered page state can change independently of the pending queue, and polling may complete before the initial or paginated page request.

## Fix

Keep the existing polling response, timestamp cursor, pagination, events and cache behavior unchanged.

Deduplication now happens only when building the pending-new queue:

* track logical entry IDs that have actually been delivered to the client
* suppress only `type: "new"` entries whose logical ID is already known
* collapse repeated `new` IDs within the same polling response
* preserve a late, previously unseen `new` entry regardless of timestamp
* do not treat an unseen polling `update` or `delete` as proof that the original `new` was already delivered
* when a page-load response completes after polling, remove from the pending queue only IDs actually present in that rendered response

Paged responses are flattened server-side, so a rendered `update` also establishes that its logical entry ID is known to the client.

The bucketed polling URLs and cache cohorts remain unchanged.

## Tests

Adds unit coverage for:

* known-ID duplicate `new` entries
* repeated IDs within one polling response
* late `new` entries after unseen polling `update`/`delete`
* rendered updated entries
* polling-before-page-load races
* preserving unrelated pending entries
* advancing polling metadata even when a duplicate is not queued
* malformed entries not poisoning the known-ID lookup

## Validation

The original issue was reproduced consistently on a production WordPress liveblog.

Before the fix, the first polling cycle could show `1 new entry available` without any new publication.

After applying equivalent ID-based deduplication:

* no false update nag is shown
* genuine new entries auto-render normally when the latest entry is visible
* genuine new entries correctly produce the update nag when the reader is scrolled away
* loading the pending entry clears the nag without it being re-queued
* a late, previously unseen `new` entry is not silently discarded based on timestamp ordering

### DIFF
--- a/src/react/epics/polling.js
+++ b/src/react/epics/polling.js
@@ -24,6 +24,7 @@ import { polling as pollingApi, getEntries } from '../services/api';
 import { scrollToEntry } from '../actions/userActions';
 
 import { shouldRenderNewEntries } from '../utils/utils';
+import { filterRenderedNewEntries } from '../utils/polling';
 
 const startPollingEpic = ( action$, state$ ) =>
 	action$.pipe(
@@ -37,16 +38,24 @@ const startPollingEpic = ( action$, state$ ) =>
 						state$.value.config
 					).pipe(
 						timeout( 10000 ),
-						map( ( res ) =>
-							pollingSuccess(
-								res.response,
+						map( ( res ) => {
+							const response = {
+								...res.response,
+								entries: filterRenderedNewEntries(
+									res.response.entries,
+									state$.value.api.entries
+								),
+							};
+
+							return pollingSuccess(
+								response,
 								shouldRenderNewEntries(
 									state$.value.pagination.page,
 									state$.value.api.entries,
 									state$.value.polling.entries
 								)
-							)
-						),
+							);
+						} ),
 						catchError( ( error ) => of( pollingFailed( error ) ) )
 					)
 				)
--- a/src/react/epics/polling.js
+++ b/src/react/epics/polling.js
@@ -24,7 +24,7 @@ import { polling as pollingApi, getEntries } from '../services/api';
 import { scrollToEntry } from '../actions/userActions';
 
 import { shouldRenderNewEntries } from '../utils/utils';
-import { filterRenderedNewEntries } from '../utils/polling';
+import { filterKnownNewEntries } from '../utils/polling';
 
 const startPollingEpic = ( action$, state$ ) =>
 	action$.pipe(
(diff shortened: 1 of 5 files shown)


GaryJones:
One question on the deliberate "remove entries older than the current polling baseline" rule in `filterKnownNewEntries` (`entryTimestamp < newestTimestamp`).

Because `newestEntry` can be an `update`/`delete` whose `timestamp` is the *edit* time of the new backing comment (per `class-wpcom-liveblog-entry.php` `for_json`: `timestamp => get_timestamp()`, while `id` is the replaced comment), the baseline timestamp can sit well above the newest genuine `new` entry. The poll request window already starts at `bucket(newestTimestamp)`, so the filter can only drop entries in the narrow `[bucketStart, newestTimestamp)` sub-bucket window — but within that window a genuinely-new, never-rendered entry delivered late (CDN cache split from its co-bucket neighbours) would now be dropped rather than queued/rendered as it was before this change.

I think this is an acceptable trade for a live blog (you generally do not back-fill entries older than what is already shown), and it looks intentional given the PR description. Just wanted to confirm you had the `update`/`delete`-baseline case in mind and are happy that a late-delivered new entry in that sub-bucket window is silently dropped. If so, no change needed — the fix and its `update`/`delete` carve-out otherwise look correct and well-tested.

lstamellos:
@GaryJones : Thanks! That was a very useful catch.

I reconsidered and decided not to keep the timestamp-based suppression. Since `newestEntry` can represent an update/delete backing comment, the fix is now revised so that timestamp ordering remains responsible only for the existing polling cursor/cache behavior, while deduplication is limited to the pending-new queue and based on explicit knowledge of logical entry IDs.

The revised approach:

* leaves the original polling response untouched for the API, events, pagination and `newestEntry` reducers;
* tracks IDs actually delivered to the client;
* does not treat an unseen polling `update` or `delete` as proof that the original `new` was already delivered;
* also handles the race where polling completes before the initial page-load response;
* keeps the existing bucketed polling/cache behavior unchanged.

So a late, previously unseen `new` entry is preserved regardless of timestamp, while an entry whose logical ID is demonstrably already known is not re-queued.

Thanks again for flagging the update/delete-baseline case. It exposed that the timestamp comparison was doing more than deduplication should.

